PR #6, opened 2021-09-09
Mm/create pypi package

MatthieuMoreau0:
- Adjust `setup.py` file for proper packaging
- Setup github actions to publish the package automatically to PYPI Test and PYPI. The package is published to PYPI only if the commit is tagged.


README.md
```diff
@@ -1,3 +1,68 @@
 # Cloud SDK
 
 SDK to be used to access Pasqal Cloud Services.
+
+## Installation
+
+To install the latest release of the `cloud-sdk`, have Python 3.8.0 or higher installed, then use pip:
```
bejito, line 7: Would be worth testing this assumption. We use pretty basic features, I'm sure we also support lower versions of python. 
Not in the scope of this PR though

README.md
```diff
@@ -1,3 +1,68 @@
...
+client_id="your_client_id" # Replace this value by the client id of your API key
+client_secret="your_client_secret" #Replace this value by the client secret of your API key
+
+sdk = SDK(client_id=client_id, client_secret=client_secret)
+
+# Creation of a sequence
+reg = Register.square(2, prefix="q")
+sequence = Sequence(reg, devices.Chadoq2)
+## You should add channels and variables to your sequence
+## See Pulser documentation for details
```
bejito, line 54: nit: double #

.github/workflows/publish-to-test-pypi.yml
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+  build-n-publish:
+    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@master
+    - name: Set up Python 3.9
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.9
```
bejito, line 18: We should use the minimal version we say we support, so 3.8 for now

setup.py
```diff
@@ -18,19 +18,22 @@
 exec(open("sdk/_version.py").read())
 
 setup(
-    name="sdk",
-    version="0.0.1",
+    name="pasqal-sdk",
+    version="0.0.3",
```
bejito, line 22: When publishing a new version to pypi, is this the version number taken into account ? I.e. does this mean we nede to manually bump this when we want to publish a new version ?

.github/workflows/publish-to-test-pypi.yml
```diff
@@ -0,0 +1,40 @@
+name: Publish cloud-sdk 📦 to PyPI and TestPyPI
+on: 
+  push:
+    branches:
```
bejito, line 4: We publish to pypi everytime we merge in dev or master ?

README.md
MatthieuMoreau0: We actually don't support python 3.7 because of the typings module. For example, `Literal` is new in python 3.8 (see (here)[https://docs.python.org/3/library/typing.html#typing.Literal] ) and we use it for our endpoints. I initially tried it with python 3.7 and I couldn't use the SDK.

It looks like on pulser they added this to their `setup.py` : 

```
    extras_require={
        ":python_version == '3.7'": [
            "backports.cached-property",
            "typing-extensions",
        ],
    },
```

I am wondering if this fixes the issue

setup.py
MatthieuMoreau0: Right now, yes. I think we could/should use the `__version__` in `sdk/_version.py` instead like in the pulser repo.

I'll take a look at it in this PR

.github/workflows/publish-to-test-pypi.yml
MatthieuMoreau0: I feel like we'll need regular changes to the library for our other services and we'll need to publish the new package for the changes to take effect. It can be cumbersome to have to merge everything to dev then create a PR from dev to master for minimal changes (e.g; my one line changes in the last PRs).

As soon as the SDK is "stable" we'll publish only on the master branch to PYPI. 


Thinking out loud, maybe we could publish to Test PyPI repository on dev and PYPI on master and add an env variable to let the end-to-end testing framework chose the package (and repository) it wants to pull. I think that could be pretty cool later on but that's a bit overkill now since we are the only user of the SDK for now.

README.md
bejito: ok, nvm

.github/workflows/publish-to-test-pypi.yml
bejito: I have never used the testpypi... Your plan sounds reasonable to me. Publish from dev for now, and from master later on.
In that case, should we disable publishing from master for now ?

bejito (approved):

PR #9, opened 2021-12-21
Fix/ Use info route to fetch group id

MatthieuMoreau0:
API Keys permissions have been restricted and it is no longer possible to retrieve the list of api keys using a given api key, so this endpoint cannot be queried to retrieve the group_id. Use the more natural /auth/info instead


bejito:
LGTM :trollface:

PR #11, opened 2022-01-20
Fix/ Add option to fetch job results in get_jobs

MatthieuMoreau0:
Results are no longer exposed in the API when fetching a job list to save bandwidth. They need to be retrieved by querying each job successively. 
I added an option to fetch the results in the _get_jobs method. By default when a user waits for the results after declaring a batch as complete, fetch the job results.

### DIFF
--- a/sdk/_version.py
+++ b/sdk/_version.py
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.0.8"
+__version__ = "0.0.9"
--- a/sdk/__init__.py
+++ b/sdk/__init__.py
@@ -57,11 +57,12 @@ class SDK:
         self.batches[batch.id] = batch
         return batch
 
-    def get_batch(self, id) -> Batch:
+    def get_batch(self, id: int, load_results: bool = False) -> Batch:
         """Retrieve a batch's data and all its jobs.
 
         Args:
             id: Id of the batch.
+            load_results: whether to load job results
 
         Returns:
             Batch: the batch stored in the PCS database.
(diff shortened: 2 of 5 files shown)


sdk/_version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.0.8"
+__version__ = "0.0.9"
```
anneclairelh, line 15: Nit out of curiosity, which kind of changes trigger a version update ?

MatthieuMoreau0: Currently a github action is set up to release a new version of the package every time the dev branch is updated (i.e. after each merge request.

So anytime we want to release a new version of the SDK we manually bump the version here, and it will automatically get built and released when merged. 

When the SDK is more stable we will release a new package version more sporadically. We will probably move the github action to the master branch, and only bump version when merging the dev branch onto master.

MatthieuMoreau0:
@anneclairelh Not in this repo no. We don't have unit tests set up. The only thing we could test for is that the proper API calls are performed for each function of the sdk, but I feel like the added value is pretty low.. I could be wrong though!

We have a couple sdk tests in our e2e test framework which ensure the basic functionalities of the sdk and its integration with the rest of the stack. For now it seems to be doing a good job catching bugs and regression.

anneclairelh:
Ok, SGTM.
Nit: Is the SDK part of our "cloud release 2022"? ( I guess so) If so, what about trying to add some unit tests there before that release if it is not too cumbersome ?

MatthieuMoreau0:
I think so yeah, I'll add a ticket in the backlog to add some unit tests.

PR #12, opened 2022-01-24
Feat/ Replace explicit device_type for batch creation by optional emulator flag

MatthieuMoreau0:
Pasqal Cloud Services do not expect anymore the device type to create a batch. It is now extracted automatically from the serialized pulser sequence. An optional emulator flag gives the ability to the user to select to run the batch on an emulator instead of a real QPU.

- Feat/ Replace explicit device_type for batch creation by optional emulator flag
- Doc/ Update basic usage in Readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ pip install -e .
 ```
 
 Bear in mind that this installation will track the contents of your local
-cloud_sdk repository folder, so if you checkout a different branch (e.g. ``master``),
+cloud_sdk repository folder, so if you checkout a different branch (e.g. `master`),
 your installation will change accordingly.
 
 ### Development Requirements (Optional)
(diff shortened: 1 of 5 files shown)


sdk/endpoints.py
```diff
@@ -15,8 +15,8 @@
 from dataclasses import dataclass
 from typing import Final
 
-CORE_API_URL: Final[str] = "https://apis.dev.pasqal.cloud/core"
-ACCOUNT_API_URL: Final[str] = "https://apis.dev.pasqal.cloud/account"
+CORE_API_URL: Final[str] = "http://localhost:4200"
```
anneclairelh, line 18: Is it an equivalent ? if not what triggered this change ?

MatthieuMoreau0: No it's an oversight, this sould have not been commited I'll revert it.

anneclairelh (approved):
LGTM

PR #16, opened 2022-02-02
WIP : Feat/ Add unit tests for SDK functions

anneclairelh:
This PR adds unit tests for SDK functions. The scope is limited to the SDK functions only and does not cover the compatibility with PCS public APIs, which is covered by end-to-end tests.


sdk/__init__.py
```diff
@@ -23,8 +23,8 @@ class SDK:
 
     def __init__(
         self,
-        client_id: str,
-        client_secret: str,
+        client_id: str = "",
```
MatthieuMoreau0, line 26: I would vouch against turning these args as kwargs because that make them sound optional, which they are not

tests/test_batch.py
```diff
@@ -0,0 +1,71 @@
...
+
+import pytest
+
+from sdk import SDK
+from sdk.batch import Batch
+
+
+class TestBatch:
+    @pytest.fixture(autouse=True)
+    @patch("sdk.Client")
```
MatthieuMoreau0, line 11: Discussed offline: instead of mocking whole Client, mock only the `requests` part. Create fixtures folder to gather in the same place all responses of the API.

Maybe we could setup a script to automatically update the fixtures, not sure.

tests/test_batch.py
```diff
@@ -0,0 +1,71 @@
...
+            id=1,
+            user_id=1,
+            priority=1,
+            status="",
+            webhook="",
+            _client=mock_client,
+            sequence_builder=self.pulser_sequence,
+        )
+        self.batch.declare_complete(wait=True)
+        self.batch._client._complete_batch.assert_called_once()
```
MatthieuMoreau0, line 71: assert that job result have been fetched

Add test with wait=False (job results not fetched)

MatthieuMoreau0 (commented):
Definitely a step in the right direction. Eventually we will need to test the client as well, particularly the login and retry logic.

sdk/__init__.py
anneclairelh: Done

tests/test_batch.py
anneclairelh: Done

tests/test_batch.py
anneclairelh: Done

sdk/client.py
```diff
@@ -134,6 +134,7 @@ def _get_jobs(self, batch_id: int, fetch_results: bool = False) -> List:
         # Fetch them explicitly if wanted
         if fetch_results:
             for job_data in job_list:
+                print(job_data)
```
MatthieuMoreau0, line 137: rm

tests/fixtures/api/auth/login/_.POST.json
```diff
@@ -0,0 +1 @@
+{}
```
MatthieuMoreau0, line 1: Weird that the response body is empty, there should be a token I think ?

tests/test_batch.py
```diff
@@ -1,71 +1,91 @@
...
+
+
 class TestBatch:
     @pytest.fixture(autouse=True)
-    @patch("sdk.Client")
-    def init_sdk(self, mock_client):
-        self.sdk = SDK()
-        self.pulser_sequence = "pulser_test_sequence"
+    @requests_mock.Mocker(kw="mock")
+    def init_sdk(self, **kwargs):
```
MatthieuMoreau0, line 19: nit: I would deconstruct the "mock" kwarg from the dictionnary

`def init_sdk(self, mock=None, **kwargs)`

tests/test_batch.py
```diff
@@ -1,71 +1,91 @@
...
+JSON_FILE = "_.{}.json"
+
+
 class TestBatch:
     @pytest.fixture(autouse=True)
-    @patch("sdk.Client")
-    def init_sdk(self, mock_client):
-        self.sdk = SDK()
-        self.pulser_sequence = "pulser_test_sequence"
+    @requests_mock.Mocker(kw="mock")
```
MatthieuMoreau0, line 18: nit: it might be useful to add this fixture to the `conftest.py` to benefit from this in other future test files

tests/test_batch.py
```diff
@@ -1,71 +1,91 @@
...
-            status="",
-            webhook="",
-            _client=mock_client,
-            sequence_builder=self.pulser_sequence,
-        )
-        job = self.batch.add_job(runs=50, variables={"var": 2})
-        self.batch._client._send_job.assert_called_once()
-        assert job.batch_id == batch_id
+    def test_create_batch(self):
+        with self.requests_mock:
```
MatthieuMoreau0, line 40: question : Do you need the context manager when using the mock decorator ?

tests/test_batch.py
```diff
@@ -1,71 +1,91 @@
...
+            batch = self.sdk.create_batch(
+                serialized_sequence=self.pulser_sequence,
+            )
+            job = batch.add_job(
+                runs=self.n_job_runs,
+                variables={"Omega_max": 14.4, "last_target": "q1", "ts": [200, 500]},
+            )
+            assert job.batch_id == self.batch_id
+            assert job.runs == self.n_job_runs
```
MatthieuMoreau0, line 57: These tests are a bit artificial since we just check the value against the one in the fixture ut not sure we have anything meaningful to test anyway. At least we will detect exceptions.

tests/test_batch.py
```diff
@@ -1,71 +1,91 @@
...
+
+    def test_batch_add_job_and_wait_for_results(self):
+        with self.requests_mock:
+            batch = self.sdk.create_batch(
+                serialized_sequence=self.pulser_sequence,
+            )
+            job = batch.add_job(
+                runs=self.n_job_runs,
+                variables={"Omega_max": 14.4, "last_target": "q1", "ts": [200, 500]},
+                wait=True,
```
MatthieuMoreau0, line 67: maybe we could assert that we called the GET /jobs/id route (I assume this is possible with request_mock) to retrieve the job results

tests/test_batch.py
```diff
@@ -1,71 +1,91 @@
...
+    def test_batch_declare_complete_and_wait_for_results(self):
+        job_id = 22010
+        with self.requests_mock:
+            batch = self.sdk.create_batch(
+                serialized_sequence=self.pulser_sequence,
+            )
+            rsp = batch.declare_complete(wait=True)
+            assert rsp["complete"]
+            assert batch.jobs[job_id].batch_id == self.batch_id
+            assert batch.jobs[job_id].result == self.job_result
```
MatthieuMoreau0, line 91: Same remark here we could assert that we fetched the job results

MatthieuMoreau0 (commented):
This is great 

For this PR:
- see minor comments
- Would be nice to run the unit tests in the CI

For another PR:
- we should add unit tests for the SDK client as well

anneclairelh (commented):
I created fixtures manually copying the ones from User Portal. I'll ask Benjamin how he created them. If manually as well, we can address the automation in this ticket: http://gitlab.local:30080/pcs/public-apis/-/issues/89 

Follow-up ticket for client unit tests:  http://gitlab.local:30080/pcs/public-apis/-/issues/88

tests/test_batch.py
MatthieuMoreau0: No I don't think the request args should appear in json. However, just thinking about it, what I think we could test here is actually test the body of the request that was sent, e.g. making sure the group_id was automatically set in the request body!

That's one of the "main" feature of this SDK so that's precisely what we should test

Same thing for job creation, make sure the batch id is properly set for the user

tests/test_batch.py
MatthieuMoreau0: Yes, agreed. But if for some reason we added the result to the fixture by mistake, the assertion could have been verified even if we didn't properly fetch the results explicitly. Since fixtures are a bit artificial (and probably subject to errors / risk not being up to date) I think  we should not rely solely on them and testing the actual request is a good fixture-agnostic way to do that imo

tests/conftest.py
```diff
@@ -0,0 +1,31 @@
...
+        )
+        with open(json_path) as json_file:
+            return json.load(json_file)
+
+    # Configure requests to use the local JSON files a response
+    mock.register_uri(requests_mock.ANY, requests_mock.ANY, json=mock_response)
+    return mock
+
+
+@pytest.fixture(scope="class")
```
MatthieuMoreau0, line 27: Question : I see you defined the scope of the request mock to "class". I assume this is the reason why you have to explictly call the reset method in the test class initialization function. Any reason why you did not define with a "function" scope which I guess should take care of resetting it automatically for you ? I'm not sure I see any benefit of keeping the same request_mock between two separate tests.

MatthieuMoreau0 (approved):
One question but looks good to me ! :+1:

tests/conftest.py
anneclairelh: Done

MatthieuMoreau0:
LGTM :rocket:

PR #17, opened 2022-03-01
Feat / update sdk for iroise

anneclairelh:
This MR updates the SDK for Iroise (following the `public-apis` [update](http://gitlab.local:30080/pcs/public-apis/-/merge_requests/109)) by:
- adding a required `jobs` argument to the `create_batch` function
- adding an optional `wait` argument to that same function in case the user wants to wait for the batch results
- updating/adding tests for new functionalities
- skipping tests for temporarily disabled functionalities (`add_job` and `complete_batch` which won't be usable since the batch will automatically be completed).

### DIFF
--- a/sdk/__init__.py
+++ b/sdk/__init__.py
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from sdk.batch import Batch
+import time
+from typing import List
+
+from sdk.batch import Batch, RESULT_POLLING_INTERVAL
 from sdk.client import Client
 from sdk.endpoints import Endpoints
 from sdk.job import Job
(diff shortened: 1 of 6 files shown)


sdk/__init__.py
```diff
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from sdk.batch import Batch
+import time
```
MatthieuMoreau0, line 15: unused import ?

sdk/client.py
```diff
@@ -104,11 +104,13 @@ def _request(self, method: str, url: str, payload: Dict = None):
 
     def _send_batch(self, batch_data: Dict):
```
MatthieuMoreau0, line 105: nit: Can you type the return value ?

sdk/job.py
```diff
@@ -28,3 +28,8 @@ class Job:
     errors: List[str]
     result: str = None
     variables: Dict = None
+
+    def __init__(self, runs, **kwargs):
```
MatthieuMoreau0, line 32: why is runs the only positionnal / mandatory argument ?

sdk/__init__.py
anneclairelh: It is actually, when we wait until each try to pull the results.

sdk/job.py
anneclairelh: Because it seemed to me we allowed jobs to be created without `variables`. To put it differently, it is the only non-nullable `Job` field that is used by the user to create the Job (before submitting it), the other required fields are filled by the API calls when actually submitting the job. Would you agree with that ?

MatthieuMoreau0: Seen offline: use dictionnary to create jobs instead of partial Job object

MatthieuMoreau0:
Don't forget to bump the version!

sdk/job.py
anneclairelh: Done

sdk/client.py
anneclairelh: Done

MatthieuMoreau0 (approved):
LGTM! We need this asap so let's get this merged. We should create a ticket in the backlog to update the documentation in the readme to match the new APIs before the release of the MVP

anneclairelh:
http://gitlab.local:30080/pcs/public-apis/-/issues/100 for updating the documentation.

PR #18, opened 2022-04-07
Feat/ Retrieve job results as dict instead of string

MatthieuMoreau0:
Job results are now sent in a dictionnary instead of in a string. Update the job class accordingly.

A new API endpoint has been defined to retrieve all job results for a given batch. Use that endpoint to retrieve job results when asked by the user instead of making a request for each job separately

### DIFF
--- a/sdk/client.py
+++ b/sdk/client.py
@@ -137,9 +137,11 @@ class Client:
         # Job results are not included when fetching a list of jobs
         # Fetch them explicitly if wanted
         if fetch_results:
+            results = self._request(
+                "GET", f"{self.endpoints.core}/api/v1/batches/{batch_id}/result"
+            )["data"]
             for job_data in job_list:
-                if job_data.get("status") == "DONE":
-                    job_data["result"] = self._get_job(job_data["id"]).get("result")
+                job_data["result"] = results.get(job_data["id"], None)
         return job_list
 
     def _get_job(self, job_id: int) -> Dict:
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ batch = sdk.create_batch(serialized_sequence)
 job1 = batch.add_job(runs=400)
 job2 = batch.add_job(runs=100, wait=True) # You can wait for the job results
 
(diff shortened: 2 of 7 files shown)


sdk/client.py
```diff
@@ -137,9 +137,11 @@ def _get_jobs(self, batch_id: int, fetch_results: bool = False) -> List:
         # Job results are not included when fetching a list of jobs
         # Fetch them explicitly if wanted
         if fetch_results:
+            results = self._request(
+                "GET", f"{self.endpoints.core}/api/v1/batches/{batch_id}/result"
```
anneclairelh, line 141: Isn't it `/results` ?

sdk/client.py
```diff
@@ -137,9 +137,11 @@ def _get_jobs(self, batch_id: int, fetch_results: bool = False) -> List:
...
+            )["data"]
             for job_data in job_list:
-                if job_data.get("status") == "DONE":
-                    job_data["result"] = self._get_job(job_data["id"]).get("result")
+                job_data["result"] = results.get(job_data["id"], None)
```
anneclairelh, line 144: I guess you're making this change so that all results are fetched at once ? Do you agree that in theory what was done before would still work ?
If so, shouldn't we also change the way it's done in `create_batch` when we wait for the results ? (indeed, that function still gets job results by calling `_get_job` for each job). 
If we do that, we'll probably need to update the batch fixture with some results, so that the test passes.

sdk/client.py
MatthieuMoreau0: Yes good catch!

sdk/client.py
MatthieuMoreau0: Yeah the goal is to make a single API call instead of one for every job. In the `create_batch`function we need to retrieve all the jobs data and not only the results so I think we can't really change the implementation with something like this.

anneclairelh: Ok!

anneclairelh (approved):
LGTM!

PR #20, opened 2022-04-11
Fix  / Device status for batch

anneclairelh:
This puts the `device_status` as a batch and not job attribute.
See failing e2e test: http://gitlab.local:30080/pcs/cloud-e2e-tests/-/pipelines/5501


MatthieuMoreau0:
My bad! Could you add `device_status` description in the docstring of the `Batch` dataclass ?

MatthieuMoreau0 (approved):

PR #25, opened 2022-07-13
Fix / Get batch function

anneclairelh:
The `get_batch` function with `load_results=True` was not working properly due to the attribute `jobs` of the `batch` object being assigned sometimes a list and sometimes a dict.
This fixes it, and also rationalizes the client functions so that there is only one way to get the batch results at once.
The end-to-end tests will be modified to include the use case that triggered this error.

### DIFF
--- a/sdk/__init__.py
+++ b/sdk/__init__.py
@@ -72,7 +72,7 @@ class SDK:
         if wait:
             while batch_rsp["status"] in ["PENDING", "RUNNING"]:
                 time.sleep(RESULT_POLLING_INTERVAL)
-                batch_rsp = self._client._get_batch(batch.id)
+                batch_rsp, _ = self._client._get_batch(batch.id)
             for job_id, job in batch.jobs.items():
                 job_rsp = self._client._get_job(job_id)
                 batch.jobs[job.id] = Job(**job_rsp)
@@ -89,12 +89,9 @@ class SDK:
             Batch: the batch stored in the PCS database.
         """
 
-        batch_rsp = self._client._get_batch(id)
+        batch_rsp, jobs_rsp = self._client._get_batch(id, fetch_results=load_results)
         batch = Batch(**batch_rsp, _client=self._client)
-
-        job_rsp = self._client._get_jobs(id, fetch_results=load_results)
-        for job in job_rsp:
-            batch.jobs[job["id"]] = Job(**job)
-
+        for job_rsp in jobs_rsp:
+            batch.jobs[job_rsp["id"]] = Job(**job_rsp)
         self.batches[batch.id] = batch
         return batch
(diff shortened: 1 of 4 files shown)


sdk/__init__.py
```diff
@@ -89,12 +89,9 @@ def get_batch(self, id: int, load_results: bool = False) -> Batch:
             Batch: the batch stored in the PCS database.
         """
 
-        batch_rsp = self._client._get_batch(id)
+        batch_rsp, jobs_rsp = self._client._get_batch(id, fetch_results=load_results)
```
MatthieuMoreau0, line 92: Should we standardize `load_results` and `fetch_results` or are they any different ?

sdk/client.py
```diff
@@ -125,24 +125,19 @@ def _send_job(self, job_data: Dict):
...
         ]
-
-    def _get_jobs(self, batch_id: int, fetch_results: bool = False) -> List:
-        job_list = self._request(
-            "GET", f"{self.endpoints.core}/api/v1/jobs?batch_id={batch_id}"
-        )["data"]
-        # Job results are not included when fetching a list of jobs
-        # Fetch them explicitly if wanted
+        jobs_data = batch_data.pop("jobs", {})
         if fetch_results:
```
MatthieuMoreau0, line 133:
```suggestion
        if fetch_results:
          results = self._request(
                  "GET", f"{self.endpoints.core}/api/v1/batches/{id}/results"
              )["data"]
          for job_data in jobs_data:
                  job_data["result"] = results.get(job_data["id"], None)
```

remove duplicate if

sdk/batch.py
```diff
@@ -91,9 +91,9 @@ def declare_complete(self, wait: bool = False) -> Dict:
         if wait:
             while rsp["status"] in ["PENDING", "RUNNING"]:
                 time.sleep(RESULT_POLLING_INTERVAL)
-                rsp = self._client._get_batch(self.id)
-            jobs_rsp = self._client._get_jobs(self.id, fetch_results=True)
-            for j in jobs_rsp:
-                job = Job(**j)
+                rsp, _ = self._client._get_batch(self.id)
+            for job_id, job in self.jobs.items():
```
MatthieuMoreau0, line 95: why not use `fetch_results` arg from `_get_batch` here ?

MatthieuMoreau0 (commented):
Is the PR to test against this bug in e2e tests up ?

MatthieuMoreau0 (approved):
LGTM! Don't forget to bump the version of the SDK for pypi

sdk/batch.py
anneclairelh: > LGTM! Don't forget to bump the version of the SDK for pypi

I did it already right ? :)

MatthieuMoreau0: Yes sorry, I wasn't looking at all the changes! lgtm

MatthieuMoreau0:
This PR is needed for some internal testing, so I will merge this. We'll iterate on top if we figure out new issues

PR #26, opened 2022-07-13
Feat / Added different emulator types when creating a batch

madagra:
This is a proposal (so feel free to reject it) for introducing new types of emulators when creating a batch. Notice that it introduces a small API change in the batch creation method but it should be very impactful.


anneclairelh:
Maybe let's update the tests of this folder so we try various types of device types ? (or at least we should make sure the tests still pass with the default value, ie `DeviceType.QPU` )

madagra:
> Maybe let's update the tests of this folder so we try various types of device types ? (or at least we should make sure the tests still pass with the default value, ie `DeviceType.QPU` )

Good point. I will do that.

sdk/__init__.py
```diff
@@ -48,19 +55,26 @@ def create_batch(
...
+        emulator = (
+            device_type == DeviceType.CPU_EMULATOR
+            or device_type == DeviceType.GPU_EMULATOR
+        )
+
         batch_rsp, jobs_rsp = self._client._send_batch(
             {
                 "sequence_builder": serialized_sequence,
                 "emulator": emulator,
+                "device_type": device_type,
```
MatthieuMoreau0, line 77: I feel like havng both these "emulator" and "device_type" fields is a bit redundant. I would rather keep only the "emulator" field and replace the boolean value by an StrEnum value 

On the API side we would check whether the emulator field is in the request body: 
- if yes, set the device_type to that value (insteadof hardcoded emulator right now)
- if not extract the device_type from the pulser sequence (already done)

Thoughts ?

madagra: @MatthieuMoreau0 Yes, that's a good idea. It might also lead to slight simplification of the code in the backend.

madagra:
@anneclairelh @MatthieuMoreau0 I addressed the comments and slightly updated the unit tests to consider multiple devices.

However, I realized that with the way tests are currently implemented, it was a bit difficult to add a second POST call to the `/api/v1/batches/` endpoint to create one batch with a different device. Let me know if I am missing something.

sdk/__init__.py
```diff
@@ -48,23 +55,30 @@ def create_batch(
...
+
+        req = {
+            "sequence_builder": serialized_sequence,
+            "webhook": self.webhook,
+            "jobs": jobs,
+        }
+
+        # the emulator field is only added in the case
+        # an emulator job is requested otherwise it's left empty
+        if is_emulator:
```
anneclairelh, line 78: nit since `is_emulator` is only used here, either I would directly put : `if device_type != DeviceType.QPU` here or at least I would move the definition right above.

tests/test_batch.py
```diff
@@ -42,6 +45,10 @@ def test_create_batch_and_wait_for_results(self, request_mock):
             == f"{self.sdk._client.endpoints.core}/api/v1/jobs/{self.job_id}"
         )
         assert batch.jobs[self.job_id].result == self.job_result
+        job = {"runs": self.n_job_runs, "variables": self.job_variables}
+        batch = self.sdk.create_batch(
+            serialized_sequence=self.pulser_sequence, jobs=[job], wait=True
+        )
```
anneclairelh, line 51: Here what are you trying to test exactly by adding these lines? Maybe we should make a separate test if we are testing another functionality ?

anneclairelh:
> @anneclairelh @MatthieuMoreau0 I addressed the comments and slightly updated the unit tests to consider multiple devices.
> 
> However, I realized that with the way tests are currently implemented, it was a bit difficult to add a second POST call to the `/api/v1/batches/` endpoint to create one batch with a different device. Let me know if I am missing something.

I'm not sure I understand what you mean exactly ? Thanks a lot for the changes :)

madagra:
> > @anneclairelh @MatthieuMoreau0 I addressed the comments and slightly updated the unit tests to consider multiple devices.
> > However, I realized that with the way tests are currently implemented, it was a bit difficult to add a second POST call to the `/api/v1/batches/` endpoint to create one batch with a different device. Let me know if I am missing something.
> 
> I'm not sure I understand what you mean exactly ? Thanks a lot for the changes :)

I meant that the way tests are currently designed does not make it easy to execute batches with different parameters since we are always pointing to the same `_POST.json` file when mocking the response to a `create_batch` request. I might be wrong since I am not familiar with this kind of mocking, but I feel that we might want to refactor the tests a bit here. Hope is clearer, let me know if I missed something.

sdk/__init__.py
madagra: Good point :)

tests/test_batch.py
madagra: Ops, forgot it. It should not be there. I was trying to find a way for mocking a second batch ( with `id` = 2 and different configuration), but I realized that it is not so straightforward.

madagra:
@anneclairelh @MatthieuMoreau0 I addressed all the comments, it is ready for merging to me now (and tested with the backend too).

sdk/batch.py
```diff
@@ -33,6 +33,7 @@ class Batch:
         - jobs: Dictionnary of all the jobs added to the batch.
         - jobs_count: number of jobs added to the batch
         - jobs_count_per_status: number of jobs per status
+        - slurm_job_id: The id of the job assigned by the Slurm scheduler (emulators only)
```
MatthieuMoreau0, line 36: Do we really want to expose that field to users ?

MatthieuMoreau0 (commented):
One minor comment otherwise LGTM

sdk/batch.py
madagra: @MatthieuMoreau0 There's an ongoing discussion on that side. Frankly, even though I added it, I think it's best not to expose it (it is pretty useless anyway for the users). If you agree, I am going to remove it.

MatthieuMoreau0: Yeah let's remove it! Don't forget to add it to excluded fields in the marshmallow schema in `public_apis`

awennersteen: I agree, there is no use for the user in having access to this

anneclairelh: I think it was already removed from the public schema in public-apis, right ?

madagra: @anneclairelh Yes, I confirm that this has already been done.

MatthieuMoreau0 (approved):

PR #28, opened 2022-08-25
Fix: HTTP errors with data = None

bejito:
`obj.get("data", {})` returns `None` if the `data` field is explicitly set to None. the default value of `get` is used only if the key is not found in the object.

500 errors seem to return a payload with a null data, thus failing the parsing.

### DIFF
--- a/sdk/_version.py
+++ b/sdk/_version.py
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.0.25"
+__version__ = "0.1.0"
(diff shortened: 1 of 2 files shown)


MatthieuMoreau0 (approved):
LGTM, don't forget to bump the version of the sdk before merging.

Have we created a ticket to handle the 500 errors in the public-apis ?

bejito:
No I did not.Frankly I don't remember what the error was...

bejito:
Yes of course ! 
I'll be bumping to v0.1.0, are you ready ?

PR #31, opened 2022-08-31
[FIX] Save jobs result properly in batch object

MatthieuMoreau0:
JSON keys are necessarily string, yet the jobs are indexed with their id as an integer, thus `results.get(job_data["id"])` was always returning None. This fixes it by casting the id to `str` when parsing the response of the request.

### DIFF
--- a/sdk/_version.py
+++ b/sdk/_version.py
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.0.24"
+__version__ = "0.0.25"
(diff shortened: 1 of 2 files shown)


anneclairelh (commented):
Question: how come we started to encounter this issue now ? Is there an additional test we should add?

MatthieuMoreau0:
@anneclairelh Yes we are probably lacking a test. I discovered the bug while playing around with our end-to-end tests, completing the batch manually since our APIs was no longer doing it on its own. That's where the bug appeared.

I will see if I can add a unit-test for this.

anneclairelh:
@MatthieuMoreau0 maybe let's add a task for adding a unit test and merge this now ?

anneclairelh (approved):

MatthieuMoreau0:
Ah yes, you're right I forgot about this.. I created an issue to add the unit test here: https://github.com/pasqal-io/cloud-sdk/issues/34

PR #33, opened 2022-09-08
[FIX] Make ReadMe compatible with new emulator workflow

anneclairelh:
Make users aware they should use `device_type=DeviceType.EMU_FREE`

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,8 +79,10 @@ job2 = {"runs": 50, "variables": {"omega_max": 10.5} }
 # Send the batch of jobs to the QPU and wait for the results
 batch = sdk.create_batch(serialized_sequence, [job1,job2], wait=True)
 
-# You can also choose to run your batch on an emulator using the optional argument 'emulator'
-# batch = sdk.create_batch(serialized_sequence, [job1,job2], emulator=True)
+# You can also choose to run your batch on an emulator using the optional argument 'device_type'
+# For using our free emulator, you can use DeviceType.EMU_FREE
+from sdk import DeviceType
+batch = sdk.create_batch(serialized_sequence, [job1,job2], device_type=DeviceType.EMU_FREE)
 
 # Once the QPU has returned the results, you can access them with the following:
 for job in batch.jobs.values():
--- a/sdk/_version.py
+++ b/sdk/_version.py
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.0.24"
+__version__ = "0.0.25"


MatthieuMoreau0:
Note: The documentation needs to be updated as well on docusaurus.

MatthieuMoreau0 (approved):

MatthieuMoreau0:
By the way, I'm wondering, if we need to bump the version if we are only updating the documentation.. Annoying part is that the CI will fail but releasing a new version to Pipy feels weird.

Mourad wants to release v0.1.0 so maybe it is time we enable release only on the master branch so we can package features in a single version, and do not have to bump version (without worrying about CI) everytime

README.md
```diff
@@ -79,8 +79,11 @@ job2 = {"runs": 50, "variables": {"omega_max": 10.5} }
 # Send the batch of jobs to the QPU and wait for the results
 batch = sdk.create_batch(serialized_sequence, [job1,job2], wait=True)
 
-# You can also choose to run your batch on an emulator using the optional argument 'emulator'
-# batch = sdk.create_batch(serialized_sequence, [job1,job2], emulator=True)
+# You can also choose to run your batch on an emulator using the optional argument 'device_type'
+# For using a basic single-threaded QPU emulator that can go up to 10 qubits, you can the "EMU_FREE" device type.
+# "EMU_SV", which is a more performant version, is not available on this platform yet.
```
madagra, line 84: Didn't we decide in the end to just keep EMU_FREE for now?

madagra (commented):
Small comment after Slack discussion.

anneclairelh:
> By the way, I'm wondering, if we need to bump the version if we are only updating the documentation.. Annoying part is that the CI will fail but releasing a new version to Pipy feels weird.
> 
> Mourad wants to release v0.1.0 so maybe it is time we enable release only on the master branch so we can package features in a single version, and do not have to bump version (without worrying about CI) everytime

I agree with this. Let me know if the change I just made sounds good, and then we can release v0.1.0 by merging "dev" into "master".

MatthieuMoreau0:
Souds good

MatthieuMoreau0 (approved):

README.md
awennersteen: I think SV is here also for purposes of the end to end tests

but the performant version should go live soon enough

sdk/__init__.py
```diff
@@ -25,7 +25,6 @@
 class DeviceType(StrEnum):
     QPU = "QPU"
     EMU_FREE = "EMU_FREE"
-    EMU_SV = "EMU_SV"
```
awennersteen, line 28: removing this will break e2e

madagra: @awennersteen Why don't we use `EMU_FREE` in the e2e tests?

bejito:
@MatthieuMoreau0 @anneclairelh I already merged my PR with a bump to v0.1.0

sdk/__init__.py
anneclairelh: Because "EMU_FREE" routes jobs to the Pasqman right now, versus "EMU_SV" routes them to the DGX connector which allows us to test that pathway!

MatthieuMoreau0:
@anneclairelh do we want to bump to 0.1.1 for doc update?

anneclairelh:
> @anneclairelh do we want to bump to 0.1.1 for doc update?

Ok, let's not :)

PR #36, opened 2022-09-12
FEAT: Add emulator configuration

### LINKED ISSUE #27: Add `config` field
Add a `config` field to the `SDK` class for supporting arbitrary configurations for batches.

awennersteen:
Add the base functionality for allowing users to supply extra configuration.

I will add some more to the README here based on
https://github.com/pasqal-io/cloud-sdk/pull/33

Closes
#27

### DIFF
--- a/sdk/__init__.py
+++ b/sdk/__init__.py
@@ -12,13 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from dataclasses import asdict
+import json
 import time
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional, Union
 
 from sdk.batch import Batch, RESULT_POLLING_INTERVAL
 from sdk.client import Client
 from sdk.endpoints import Endpoints
 from sdk.job import Job
+from sdk.utils.configuration import Configuration
 from sdk.utils.strenum import StrEnum
 
 
@@ -47,6 +50,7 @@ class SDK:
         serialized_sequence: str,
         jobs: List[Dict[str, Any]],
         device_type: DeviceType = DeviceType.QPU,
+        configuration: Optional[Union[Dict[str, Any], Configuration]] = None,
         wait: bool = False,
     ) -> Batch:
         """Create a new batch and send it to the API.
@@ -58,6 +62,7 @@ class SDK:
             device_type: The type of device to use, either an emulator or a QPU
               If set to QPU, the device_type will be set to the one
               stored in the serialized sequence
+            configuration: A dictionary with extra configuration for the emulators that accept it.
             wait: Whether to wait for results to be sent back
 
 
@@ -76,6 +81,12 @@ class SDK:
         if device_type != DeviceType.QPU:
             req.update({"emulator": device_type})
 
+        # The configuration field is only added in the case
+        # it's requested
+        if configuration:
+            if isinstance(configuration, Configuration):
+                configuration = asdict(configuration)
+            req.update({"configuration": json.dumps(configuration)})
         batch_rsp, jobs_rsp = self._client._send_batch(req)
         batch = Batch(**batch_rsp, _client=self._client)
         for job_rsp in jobs_rsp:
--- a/sdk/__init__.py
+++ b/sdk/__init__.py
@@ -85,7 +85,7 @@ class SDK:
         if configuration:
             if isinstance(configuration, Configuration):
                 configuration = configuration.to_dict()
-            req.update({"configuration": json.dumps(configuration)})
+            req.update({"configuration": configuration})
         batch_rsp, jobs_rsp = self._client._send_batch(req)
         batch = Batch(**batch_rsp, _client=self._client)
         for job_rsp in jobs_rsp:
--- a/sdk/__init__.py
+++ b/sdk/__init__.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from dataclasses import asdict
 import json
 import time
 from typing import Any, Dict, List, Optional, Union
@@ -85,7 +84,7 @@ class SDK:
         # it's requested
         if configuration:
             if isinstance(configuration, Configuration):
-                configuration = asdict(configuration)
+                configuration = configuration.to_dict()
             req.update({"configuration": json.dumps(configuration)})
         batch_rsp, jobs_rsp = self._client._send_batch(req)
         batch = Batch(**batch_rsp, _client=self._client)
--- a/sdk/__init__.py
+++ b/sdk/__init__.py
@@ -49,7 +49,7 @@ class SDK:
         serialized_sequence: str,
         jobs: List[Dict[str, Any]],
         device_type: DeviceType = DeviceType.QPU,
-        configuration: Optional[Union[Dict[str, Any], Configuration]] = None,
+        configuration: Optional[Configuration] = None,
         wait: bool = False,
     ) -> Batch:
         """Create a new batch and send it to the API.
@@ -83,9 +83,7 @@ class SDK:
         # The configuration field is only added in the case
         # it's requested
         if configuration:
-            if isinstance(configuration, Configuration):
-                configuration = configuration.to_dict()
-            req.update({"configuration": configuration})
+            req.update({"configuration": configuration.to_dict()})
         batch_rsp, jobs_rsp = self._client._send_batch(req)
         batch = Batch(**batch_rsp, _client=self._client)
         for job_rsp in jobs_rsp:
--- a/sdk/batch.py
+++ b/sdk/batch.py
@@ -53,7 +53,7 @@ class Batch:
     jobs: Dict[int, Job] = field(default_factory=dict)
     jobs_count: int = 0
     jobs_count_per_status: Dict[str, int] = field(default_factory=dict)
-    configuration: Optional[Union[Configuration, Dict]] = None
+    configuration: Optional[Configuration] = None
 
     def add_job(self, runs: int = 100, variables: Dict = None, wait: bool = False):
         """Add and send a new job for this batch.
--- a/sdk/batch.py
+++ b/sdk/batch.py
@@ -53,7 +53,7 @@ class Batch:
     jobs: Dict[int, Job] = field(default_factory=dict)
     jobs_count: int = 0
     jobs_count_per_status: Dict[str, int] = field(default_factory=dict)
-    configuration: Optional[Configuration] = None
+    configuration: Optional[Dict] = None
 
     def add_job(self, runs: int = 100, variables: Dict = None, wait: bool = False):
         """Add and send a new job for this batch.
--- a/sdk/batch.py
+++ b/sdk/batch.py
@@ -1,6 +1,7 @@
 import time
 from dataclasses import dataclass, field
-from typing import Dict
+from typing import Dict, Optional, Union
+from sdk.utils.configuration import Configuration
 
 from sdk.client import Client
 from sdk.job import Job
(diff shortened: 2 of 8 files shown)


sdk/utils/configuration.py
```diff
@@ -0,0 +1,10 @@
+from dataclasses import dataclass, asdict
+
+
+@dataclass
```
bejito, line 4: Too bad we can't tell python dataclasses to take any additional field in the default constructor. We could have all the Config (the static known part + the dynamic part that's left free for now) in the same object

bejito (approved):
One minor comment, but LGTM

sdk/utils/configuration.py
awennersteen: Yeah, I didn't really find any good solution that kept it all in the dataclass and would allow the user to copy-paste configuration from documentation or similar. It would have been great to be able to do that. 

Let's see if anyone has a good way of doing this.

sdk/__init__.py
```diff
@@ -76,6 +81,12 @@ def create_batch(
         if device_type != DeviceType.QPU:
             req.update({"emulator": device_type})
 
+        # The configuration field is only added in the case
+        # it's requested
+        if configuration:
+            if isinstance(configuration, Configuration):
+                configuration = asdict(configuration)
```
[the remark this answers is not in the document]
awennersteen: indeed.

sdk/utils/configuration.py
anneclairelh: I agree with this comment.
Could we save the additional parameters as a specific field (with dict type) of the dataclass "Configuration" and supercharge the "to_dict" method so that it returns a flat dictionary with all the fields ? 
Another option is to make a custom dataclass on the fly, see here: https://stackoverflow.com/questions/52534427/dynamically-add-fields-to-dataclass-objects

tests/test_batch.py
```diff
@@ -101,3 +102,18 @@ def test_batch_declare_complete_and_wait_for_results(self, request_mock):
...
         assert batch.jobs[self.job_id].batch_id == batch.id
         assert batch.jobs[self.job_id].result == self.job_result
+
+    @pytest.mark.parametrize("device_type, configuration", [
+        (DeviceType.EMU_SV, {"precision": "normal"}),
+        (DeviceType.EMU_SV, Configuration()),
+    ])
+    def test_create_batch_configuration(self, device_type, configuration):
+        job = {"runs": self.n_job_runs, "variables": self.job_variables}
+        configuration = {"precision": "normal"}
```
anneclairelh, line 112: In this case, shouldn't we make use of the dataclass and have: configuration=Configuration(precision="normal")?

sdk/utils/configuration.py
bejito: Yeah I don't think there is an actual good solution... I tried looking into `TypedDict`s, but they explicitly forbid additional keys and they don't look so user-friendly.
Do you know if we can live with a flat config, i.e. no nested fields, or are we going to need to put some dicts in the config ?

MatthieuMoreau0: Do we really need a dataclass here? The main benefit I see are for validation (which we are bothered by right now) and the automatic mapping of dict to attributes of the object, but I doubt the users are going to update a configuration object once it is created

Any major feature of dataclass we would like to benefit from ?

awennersteen: I looked at the same StackOverflow @anneclairelh, I opted against it because I thought maybe it would be less user friendly.

Wecould perhaps make a ConfigurationFactory that returns a custom Configuration. Maybe with a from_dict method too, since it might be easier to copy-paste a dictionary for a user.
It seems like over engineering though.

> Could we save the additional parameters as a specific field (with dict type) of the dataclass "Configuration" and supercharge the "to_dict" method so that it returns a flat dictionary with all the fields ?
> Do you know if we can live with a flat config, i.e. no nested fields, or are we going to need to put some dicts in the config ?

Flat should be fine for the foreseeable future. For anything that would require nested dicts configuration we can rework it in the future.

awennersteen: @MatthieuMoreau0 
Primarily I wanted something with default values, and dataclass was the first option that came to mind.

What Anne-Claire suggested re make_dataclass would not have the issue that you lost it when converting to dict. (at least based on stack overflow)

tests/test_batch.py
awennersteen: Actually this line shouldn't be here at all, not sure where it came from. 
It overrides the parameterized test... (where I wanted to use a dict just to have a test using both Configuration and Dict types)

sdk/__init__.py
```diff
@@ -76,6 +80,12 @@ def create_batch(
         if device_type != DeviceType.QPU:
             req.update({"emulator": device_type})
 
+        # The configuration field is only added in the case
+        # it's requested
+        if configuration:
+            if isinstance(configuration, Configuration):
+                configuration = configuration.to_dict()
+            req.update({"configuration": json.dumps(configuration)})
```
MatthieuMoreau0, line 88: Dump to string won't be necessary anymore if the changes from str -> JSON is made on the public_apis side

sdk/batch.py
```diff
@@ -51,6 +53,7 @@ class Batch:
     jobs: Dict[int, Job] = field(default_factory=dict)
     jobs_count: int = 0
     jobs_count_per_status: Dict[str, int] = field(default_factory=dict)
+    configuration: Optional[Union[Configuration, Dict]] = None
```
MatthieuMoreau0, line 56: Why can this be either a Configuration or a Dict object ? Can't we stick to one of the two types with proper casting on batch deserialization?

Otherwise it might be error prone when manipulating inside methods

awennersteen: Configuration doesn't apply to the QPU nor EMU_FREE.
I guess we can make it so that we have a Configuration with no fields for those and have inherited classes for the emulators with fields...

The concept of multiple Configuration objects might become reality once we put different types of emulators online.

bejito: The thing is that we want part of the config to be free-form because we're not totally sure of what that contains, but we  still wanted some structure to help users with the fields we are sure about

sdk/__init__.py
awennersteen: Done

sdk/utils/configuration.py
anneclairelh: What I feel is confusing with the current implementation is that for a user that has some configuration specifying the "precision" parameter and another parameter that is free-form, they would have no choice but pass a regular dict, so we'd loose all the benefit of the dataclass for "precision". 

If we want the dataclass uniquely for the purpose of validation, I would suggest we ask the user to provide a dictionary `configuration_dict` anyway, and then inside the create batch function we do `Configuration(**configuration_dict)` and just make sure it doesn't raise an error on the non-free fields. But then we still use `configuration_dict` to send to the cloud. The day we're set on all the configuration fields possible for all emulators, we can actually force the user to pass a `Configuration` as input.

anneclairelh: @awennersteen  any thoughts on this ?

awennersteen: For me I think that providing the users with the Configuration class providing the default values for the most commonly used parameters is more important than validation. And validation here is more for helping the user send correct requests than we relying on this in any way since these parameters are all validated in the `core` api.

It's hard to estimate percentages but I'd say the majority of users will not change the configuration, > 99% would only change the values provided and only some very few advanced users would need to supply extra fields. However, for us internally supplying extra fields will possibly be useful especially for showcasing. I wouldn't be surprised if the cloud-sdk will permanently need to support an arbitrary dictionary as configuration for some internal use.

I also expect that within a few weeks we will add more parameters to the configuration object.

awennersteen:
@anneclairelh 
I'd like to add some details to the README,
Should we merge this first and then I can push it to #33? 
Or other way around?

anneclairelh:
> @anneclairelh I'd like to add some details to the README, Should we merge this first and then I can push it to #33? Or other way around?

Just pushed a last version, if it looks good to @MatthieuMoreau0 (Matthieu, we'll take care of release flow properly later), we can merge

sdk/batch.py
MatthieuMoreau0: I don't expect users to create batch object manually, it's always going to be returned by the `get_batch` or `create_batch` function from the SDK. 

We have full control over the attributes deserialization in the batch object, which is why I find it surprising that we cannot decide for the configuration attribute to always be a `dict` or a `Configuration` instance. If configuration is too restrictive for what the backend can send back let's just use a dict for the configuration attribute here.

In the batch creation function inputs however we can be a bit more lenient and allow both types for better user experience.

sdk/utils/configuration.py
bejito: For me, the only goal of having this as a (data)class was to have type hinting for users more than having validation (that must be done in the backend in any ways).
But I agree that this ends up looking overly complicated thanks to python. The amount of discussions on something that should be simple is a good indicator.
@awennersteen what do you think of going back to a completely free-form dict ? We could make it optional ; if the user does not provide it, we just send one with default values and if the user wants to customize it, he has to enter all the values ?

awennersteen: we can do that @bejito , but personally I would prefer the current solution if I was a user.

bejito: Ok, good for me too. In that case, we don't advertise the fact that users can use a dict in the doc, just the Configuration object

sdk/batch.py
awennersteen: You made a very good point and I did change it. It now only accepts Configuration.

sdk/utils/configuration.py
```diff
@@ -0,0 +1,55 @@
...
+        conf_dict: Optional[Union[Dict[str, Any], str]], prefix: str = ""
+    ) -> Dict:
+        """flattens dictionaries, except the extra_config field.
+        Excludes None values
+        """
+        if not conf_dict:
+            return {}
+        if prefix == "extra_config":
+            #  In extra_config don't flatten the further dicts.
+            return {k: v for (k, v) in conf_dict.items()}
```
MatthieuMoreau0, line 46: Isn't this equivalent to returning conf_dict directly ?

sdk/batch.py
MatthieuMoreau0: Wonderful thanks :+1:

sdk/utils/configuration.py
awennersteen: Ahhaha yep

sdk/utils/configuration.py
```diff
@@ -0,0 +1,55 @@
...
+            for k in self.extra_config.keys():
+                if k in self.__dataclass_fields__.keys():
+                    raise InvalidConfiguration(INVALID_KEY_ERROR_MSG.format(k))
+        return self
+
+    @staticmethod
+    def _flatten_dict(
+        conf_dict: Optional[Union[Dict[str, Any], str]], prefix: str = ""
+    ) -> Dict:
+        """flattens dictionaries, except the extra_config field.
```
MatthieuMoreau0, line 39: nit: feels like this function does a bit more than it is intended which could have some side effects. I'd suggest something a bit easier to read which does only what we expect: unnest the extra_config values on the same level as the default parameters. I might have missed something though

My suggestion:
```python
def _unnest_extra_config(conf_dict) -> Dict:
    res = { k:v for (k,v) in conf_dict.items() if not k == "extra_config"}
    res.update(conf_dict["extra_config"]
    return res
```

sdk/__init__.py
```diff
@@ -76,6 +80,10 @@ def create_batch(
         if device_type != DeviceType.QPU:
             req.update({"emulator": device_type})
 
+        # The configuration field is only added in the case
+        # it's requested
+        if configuration:
+            req.update({"configuration": configuration.to_dict()})
         batch_rsp, jobs_rsp = self._client._send_batch(req)
         batch = Batch(**batch_rsp, _client=self._client)
```
MatthieuMoreau0, line 88: The result of this deserialization will lead the configuration attribute to be a dict and not a Configuration instance unless dataclasses operate some magic I'm not aware of, but it doesn't look like it in my experiments:

```bash
>>> @dataclass
... class Yolo:
...   id: int = 0
...   configuration: Configuration = None
... 
>>> 
>>> 
>>> 
>>> Yolo(**{"id": 1})
Yolo(id=1, configuration=None)
>>> Yolo(**{"id": 1, "configuration": { "dt" : 0.1}})
Yolo(id=1, configuration={'dt': 0.1})
```

With the current implementation of the configuration dataclass, I don't really see how we are going to circumvent that issue tbh. All the dict fields will be on the same level in the response of the core API and we need to separate the `extra_config` fields from the default value fields to build the configuration object.

bejito (approved):
LGTM

sdk/__init__.py
awennersteen: that was the whole point :) 

With a growing number of emulators it will be much more manageable to have them on the first level downstream.

Suppose in the future we'll have 8 different versions of the emulator, all of which will have some number of custom options and others will be shared by an arbitrary number of emulators.

To some extent, the distinction between free form and default config options are very minor, and not confirmed. Indeed, different emulators will have different default config options.

The current implementation is very hindered by the fact that this is a public repo and I can only show the parts that are for products already announced.

sdk/utils/configuration.py
```diff
@@ -0,0 +1,55 @@
+import dataclasses
```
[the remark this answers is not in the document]
awennersteen: ah right, used for a intermediate stage

anneclairelh (approved):

PR #46, opened 2022-10-25
Aw/introduce typing

### LINKED ISSUE #43: Adding mypy type checking
I'm trying to add type checking when importing CloudSdk, but it seems to be not working because CloudSdk is not fully type checked.

Are you considering to add a `.mypy.ini` file to the project and run `mypy` to fully check the project? This way, users will be able to use CloudSdk's types.

awennersteen:
Fixes #43 
Fixes unnecessarily keeping the `Configuration` as a `Dict` after sending the submit batch request.
Fixes tests fixtures

Currently not added to a pipeline, but I figured since we don't run the tests either we shouldn't?

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,8 @@
+[mypy]
+warn_return_any = True
+show_error_codes = true
+disallow_untyped_defs = true
+warn_unreachable = true
+warn_unused_ignores = true
+ignore_errors = false
+
--- a/mypy.ini
+++ b/mypy.ini
@@ -6,4 +6,6 @@ disallow_untyped_defs = true
 warn_unreachable = true
 warn_unused_ignores = true
 ignore_errors = false
-
+warn_redundant_casts = true
+allow_untyped_calls = true
+allow_any_generics = true
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,5 +1,5 @@
 [mypy]
-warn_return_any = True
+warn_return_any = true
 show_error_codes = true
 disallow_untyped_defs = true
 warn_unreachable = true
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,4 +1,5 @@
 [mypy]
+strict=true
 warn_return_any = true
 show_error_codes = true
 disallow_untyped_defs = true
--- a/sdk/batch.py
+++ b/sdk/batch.py
@@ -63,7 +63,10 @@ class Batch:
         self.configuration = Configuration.from_dict(self.configuration)
 
     def add_job(
-        self, runs: int = 100, variables: Dict = None, wait: bool = False
+        self,
+        runs: int = 100,
+        variables: Optional[Dict[str, Any]] = None,
+        wait: bool = False,
     ) -> Job:
         """Add and send a new job for this batch.
 
@@ -88,7 +91,7 @@ class Batch:
                 job = Job(**job_rsp)
         return job
 
-    def declare_complete(self, wait: bool = False) -> Dict:
+    def declare_complete(self, wait: bool = False) -> Dict[str, Any]:
         """Declare to PCS that the batch is complete.
 
         Args:
--- a/sdk/batch.py
+++ b/sdk/batch.py
@@ -6,7 +6,6 @@ from sdk.client import Client
 from sdk.job import Job
 from sdk.utils.configuration import Configuration
 
-
 RESULT_POLLING_INTERVAL = 2  # seconds
 
 
@@ -63,7 +62,9 @@ class Batch:
     def __post_init__(self) -> None:
         self.configuration = Configuration.from_dict(self.configuration)
 
-    def add_job(self, runs: int = 100, variables: Dict = None, wait: bool = False) -> Job:
+    def add_job(
+        self, runs: int = 100, variables: Dict = None, wait: bool = False
+    ) -> Job:
         """Add and send a new job for this batch.
 
         Args:
--- a/sdk/batch.py
+++ b/sdk/batch.py
@@ -1,10 +1,10 @@
 import time
 from dataclasses import dataclass, field
-from typing import Dict, Optional, Union
-from sdk.utils.configuration import Configuration
+from typing import Any, Dict, Optional
 
 from sdk.client import Client
 from sdk.job import Job
+from sdk.utils.configuration import Configuration
 
 
 RESULT_POLLING_INTERVAL = 2  # seconds
@@ -31,7 +31,7 @@ class Batch:
         - webhook: Webhook where the job results are automatically sent to.
         - sequence_builder: Pulser sequence of the batch.
         - start_datetime: Timestamp of the time the batch was sent to the QPU.
-        - end_datetime: Tiemstamp of when the  batch process was finished.
+        - end_datetime: Timestamp of when the  batch process was finished.
         - device_status: Status of the device where the batch is running.
         - jobs: Dictionary of all the jobs added to the batch.
         - jobs_count: number of jobs added to the batch.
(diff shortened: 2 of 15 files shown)


mypy.ini
```diff
@@ -0,0 +1,8 @@
+[mypy]
+warn_return_any = True
```
MatthieuMoreau0, line 2: nit: not same capitalization for True, maybe this works fine tho

MatthieuMoreau0 (approved):

sdk/batch.py
```diff
@@ -54,13 +53,18 @@ class Batch:
...
-    device_status: str = None
+    device_status: Optional[str] = None
     jobs: Dict[int, Job] = field(default_factory=dict)
     jobs_count: int = 0
     jobs_count_per_status: Dict[str, int] = field(default_factory=dict)
-    configuration: Optional[Dict] = None
+    configuration: Optional[Configuration] = None
+
+    def __post_init__(self) -> None:
+        self.configuration = Configuration.from_dict(self.configuration)
```
awennersteen, line 63: This basically let's us keep the type as Configuration, but allows us to initialize with a dict.

Very happy to receive suggestions about this

MatthieuMoreau0:
>Currently not added to a pipeline, but I figured since we don't run the tests either we shouldn't?

Definitely something we should take care of, I created an issue: https://github.com/pasqal-io/cloud-sdk/issues/new

sdk/utils/configuration.py
```diff
@@ -16,24 +18,35 @@ class Configuration:
...
+    def from_dict(
+        conf: Optional[Union[Configuration, Dict]]
+    ) -> Optional[Configuration]:
+        if conf is None:
+            return None
+        if isinstance(conf, Configuration):
+            return conf
+        dt = conf.pop("dt", Configuration.dt)
+        precision = conf.pop("precision", Configuration.precision)
+        return Configuration(dt=dt, precision=precision, extra_config=conf)
```
awennersteen, line 31: Allows the automatic conversion of various inputs to a `Configuration` object as required to initialize from a API response.

Very happy to receive suggestions on this.

sdk/client.py
```diff
@@ -103,32 +106,35 @@ def _request(self, method: str, url: str, payload: Dict = None):
...
+        self, batch_data: Dict[str, Any]
     ) -> Tuple[Dict[str, Any], List[Dict[str, Any]]]:
         batch_data.update({"group_id": self.group_id})
         batch_data = self._request(
             "POST",
             f"{self.endpoints.core}/api/v1/batches",
             batch_data,
         )["data"]
-        jobs_data = batch_data.pop("jobs", [])
+        jobs_data = batch_data.pop("jobs", {})
```
anneclairelh, line 117: Why replacing the list with a dict here?

sdk/utils/strenum.py
```diff
@@ -1,11 +1,13 @@
...
         """Used when dumping enum fields in a schema."""
-        return self.value
+        ret: str = self.value
+        return ret
 
     @classmethod
-    def list(cls):
-        return list(map(lambda c: c.value, cls))
+    def list(cls) -> List[str]:
+        return list(map(lambda c: c.value, cls))  # type: ignore
```
anneclairelh, line 13: Out of curiosity, does adding ` # type: ignore` have a concrete effect on mypy type checking ?

tests/fixtures/api/batches/_.POST.json
```diff
@@ -13,6 +13,8 @@
     "user_id": 1,
     "webhook": "10.0.1.5",
     "configuration": {"dt": 0.1, "precision": "normal"},
+    "start_datetime": "2021-11-10T15:27:06.698066",
```
anneclairelh, line 16: We can probably add this to the "jobs" fixture as well!

sdk/batch.py
anneclairelh: If `self.configuration` is a `Configuratioon` already,  what will this do?

tests/fixtures/api/batches/_.POST.json
awennersteen: Done!

sdk/batch.py
awennersteen: so I (confusingly) accept Configuration as an argument to`Configuration.from_dict` and it just returns the Configuration object.

Happy to get some ideas for how to do this better.

sdk/utils/strenum.py
awennersteen: it does. means that mypy doesn't fail on this this line, or even try to analyse it.

to make mypy work on something like this it looks like you have to split it up across several lines.

sdk/client.py
```diff
@@ -114,22 +117,25 @@ def _send_batch(
...
-            "data"
-        ]
-        jobs_data = batch_data.pop("jobs", {})
+    def _get_batch(
+        self, id: int, fetch_results: bool = False
+    ) -> Tuple[Dict[str, Any], List[Dict[str, Any]]]:
+        batch_data: Dict[str, Any] = self._request(
+            "GET", f"{self.endpoints.core}/api/v1/batches/{id}"
+        )["data"]
+        jobs_data = batch_data.pop("jobs", [])
```
awennersteen, line 138: Looks like before _get_batch and send batch had different default types.
Now they both have `List`, as they should.
Cf line 117

sdk/batch.py
MatthieuMoreau0: I'd rather check it in post_init instead of inside the `from_dict` function but that's probably a matter of taste

awennersteen: yeah, I did that in the first version, but I also thought it looked messy...
but it's also true that it's probably less confusing than accepting `None` and `Configuration` in a method called `from_dict`.

sdk/utils/configuration.py
MatthieuMoreau0: Any way we could make this programmatic ? I.e this function doesn't need to be updated if the configuration class attributes get updated. We can list all the fields using the `dataclasses.fields` function

something like
```python
from dataclasses import fields

base_conf = {}
for field in fields(Configuration):
  if field.name != "extra_config" and field.name in conf:
     base_conf[field.name]=conf.pop(field.name)
return Configuration(**base_conf, extra_config=conf
```

sdk/client.py
MatthieuMoreau0: Interesting that mypy wasn't complaining about this

awennersteen: I think it doesn't manage to understand the pop method. I tried for a time (and it might be still left in some places) to annotate the return type of what's `pop`ed. In that case it does work. But it also makes reading the code harder...

You'll see in below that in a different method here the default values of this was swapped.

anneclairelh:
@awennersteen  are you also planning to implement a CI stage running the mypy checking or should I do it in a separate issue ?

sdk/utils/configuration.py
awennersteen: Exactly the kind of configuration I was looking for. I didn't think about the `fields()`
implemented

MatthieuMoreau0: Great! Thanks for the unit test as well :)

MatthieuMoreau0 (approved):
I think enforcing the type check (and unit tests) in the CI should be the first priority for the cloud-sdk after this is merged

anneclairelh:
> @awennersteen are you also planning to implement a CI stage running the mypy checking or should I do it in a separate issue ?

If not, could you at least run the tests locally and make sure that they pass, so that we have a bug-free version of the SDK?

sdk/batch.py
awennersteen: went back on this

awennersteen:
Tests pass and mypy passes locally
I agree with Matthieu that we should do this asap but I'd like to not do it today...

I'll merge this after having had a chance to also verify in the e2e

awennersteen:
Works in e2e

PR #50, opened 2022-10-27
Extending support to Python 3.7

### LINKED ISSUE #48: Python 3.7 support
Hi!

Currently in Pulser we support Python 3.7 or higher, while for the `cloud-sdk` you are starting from Python 3.8.
As we are introducing the `cloud-sdk` as a dependency to the `pulser-pasqal` extension, it would be preferable that the supported Python versions between the two packages match.

Python 3.7 end of life is in 8 months, which puts us in a weird limbo where:
- It's a bit too soon to drop support in Pulser
- It seems useless to add support in the `cloud-sdk` since in 8 months we would be removing it

So, I think the decision here should be made based on how easy it would be to add Python 3.7 support on the `cloud-sdk`. 
- In terms of the `cloud-sdk` dependencies, I checked the supported Python versions of `requests` and `requests-mock` and there seems to be no issue there (they still support >=3.5)
- In terms of Python >=3.8 features, I wonder how much you are using. If it's just special types from typing (I found `Final` and `TypedDict`), it is very easy to workaround (we did it in Pulser). Is there anything else you can think of?

If it turns out it would be easy and you wouldn't mind, I'd be happy to take care of the Python 3.7 support, but I'd like to hear from you first.

Thanks!

HGSilveri:
See discussion in #48.

Closes #48 .

### DIFF
--- a/sdk/endpoints.py
+++ b/sdk/endpoints.py
@@ -12,8 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from sys import version_info
+
 from dataclasses import dataclass
-from typing import Final
+
+if version_info[:2] >= (3, 8):
+    from typing import Final
+else:
+    from typing_extensions import Final
 
 CORE_API_URL: Final[str] = "https://apis.pasqal.cloud/core"
 ACCOUNT_API_URL: Final[str] = "https://apis.pasqal.cloud/account"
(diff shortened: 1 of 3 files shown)


MatthieuMoreau0 (approved):
Looking good

awennersteen:
For me no problem

@MatthieuMoreau0 
Let's merge this without upgrading the version, I'll add python 3.7 to the tests and take care of the update in cloud e2e?

MatthieuMoreau0:
SGTM, just FYI the pipeline will fail @HGSilveri but that's expected since we will wait for the CI tests before pushing a new version to PYPI

PR #54, opened 2022-12-21
[FIX] return the latest batch data

awennersteen:
A fix ensuring that the `Batch` returned to the user is the latest batch object fetched, and return the results when requested. This way the status of the batch and jobs will at least be shown as DONE, and not PENDING/RUNNING.

We suggest decoupling the `wait` and `fetch_results` when retrieving a batch. `wait=True` makes the python statement blocking but doesn't fetch the results by default while fetch_results=True implies wait=True and fetches and stores in memory the job results.
The underlying idea being, if a batch has many jobs fetching all the results might not be desirable and the user might want to implement  custom logic to fetch job results, e.g. only the last job of the batch if that's the one they care about.


awennersteen:
Pending checking in e2e

tests/test_batch.py
```diff
@@ -30,22 +30,33 @@ def test_create_batch(self, device_type):
...
+        assert batch.jobs
+        for job_id, job in batch.jobs.items():
+            assert self.job_id == job_id
+            assert job.result is None
         assert request_mock.last_request.method == "GET"
-        assert (
-            request_mock.last_request.url
-            == f"{self.sdk._client.endpoints.core}/api/v1/jobs/{self.job_id}"
+
+    def test_create_batch_and_wait_for_results(self, request_mock):
```
MatthieuMoreau0, line 47: nit: for consistency with kwargs names, should this be named `test_create_batch_and_fetch_results`

MatthieuMoreau0 (commented):
Looking good ! Let's setup some e2e tests run on this.

@anneclairelh We suggest decoupling the `wait` and `fetch_result` when retrieving a batch. `wait=True` makes the python statement blocking but doesn't fetch the results by default while `fetch_results=True` implies `wait=True` and fetches and stores in memory the job results. THe underlying idea being, if a batch has many jobs fetching all the results might be costly in terms of memory and the user may want to define instead his custom logic to fetch job results, e.g. only the last job of the batch if that's the one he cares about, ... What do you think ?

If that sounds good, we should make the rest of the functions' signature consistent and add these two kwargs where needed.

awennersteen:
Thanks for taking a look Matthieu. I'll update the PR description too now that this PR had a slight scope creep...

To be clear, we'd want to add the `wait`/`fetch_results`  behavior to `Batch.add_job` and `Batch.declare_complete`.

MatthieuMoreau0 (approved):
LGTM, I will launch a battery of e2e tests with these changes for validation.

anneclairelh:
This change makes a lot of sense to me. Sorry for not reviewing it earlier! LGTM

MatthieuMoreau0:
@anneclairelh I updated the changelog let me know if we can merge this.

anneclairelh:
> @anneclairelh I updated the changelog let me know if we can merge this.

LGTM!

PR #59, opened 2023-01-31
[FEAT] Refactor Config and add EMU_FREE specific config

Augustinio:
Cf changelog for list of changes.


tests/test_batch.py
```diff
@@ -117,7 +117,10 @@ def test_batch_declare_complete_and_wait_for_results(self, request_mock):
     @pytest.mark.parametrize(
         "device_type, configuration, expected",
         [
-            (DeviceType.EMU_SV, Configuration(), Configuration()),
+            (DeviceType.EMU_SV, EmuSVConfig(), EmuSVConfig()),
+            (DeviceType.QPU, None, None),
+            (DeviceType.EMU_FREE, EmuFreeConfig(), EmuFreeConfig(extra_config={'dt': 0.1, 'precision': 'normal'})),
+            ('SomethingElse', BaseConfig(), BaseConfig(extra_config={'dt': 0.1, 'precision': 'normal'}))
```
MatthieuMoreau0, line 123: This tests look a bit weird to me, the result is not really what we would expect to happen for a user (so not really useful as a documentation tool) and the expected value is the result of an hard-coded fixture so not sure it brings much value in term of testing of the code-base. 

Do you see any value in keeping it ? Or does it need further tweaking, e.g. with more fixtures depending on the batch type, config, etc ?

MatthieuMoreau0 (commented):
Looking good !

Two minor remarks:
- For the PYPI package update, the _version file needs to be updated
- Not sure if all the device types and config should be gathered in a `utils` folder. maybe a simple device folder would beappropriate, with a configuration sub folder ? Open to suggestion!

tests/test_batch.py
Augustinio: I think it is non trivial in that we are testing the following flow:
- `create_batch` takes the `configuration` argument
- it converts it to dict
- it sends a request and feeds the response to the `Batch` model
- the batch model converts the configuration dict from the response to an instance of the proper Config class

MatthieuMoreau0: Ok I get it, it's fine I think then

MatthieuMoreau0 (approved):

PR #61, opened 2023-02-09
[FEAT] Replace EMU_SV with EMU_TN

anneclairelh:
EMU_SV is actually not finalized yet by our emulation team.
Instead, EMU_TN is now available. This updates the SDK device types accordingly.
Will open a corresponding MR in the e2e tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,14 +2,13 @@
 
 All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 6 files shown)


awennersteen:
I think we could also add `max-bond-dim` to the default parameters?
Set the default to 500

awennersteen (approved):

MatthieuMoreau0 (approved):

PR #63, opened 2023-02-27
Add group_id to the Job schema

oliver-gordon:
Group_id is being returned in Job API calls but will fail to instantiate objects without the field in the schema.


MatthieuMoreau0:
Don't forget to update the version file & the changelog (a simple bullet point will do the trick)

MatthieuMoreau0 (approved):

PR #65, opened 2023-03-02
[FEAT] Allow user to directly use a JWT token instead of logging in with username / password

anneclairelh:
This feature is needed for the Jupyter Notebook of PQD which will have the JWT token stored as an environment variable. 
Right now, it is kept as a alternative mechanism to the regular login.
In order to address the issue of token expiration, a temporary error is raised which suggests the user to perform an interactive login. Therefore, a new optional parameter is added to the SDK client, namely the `login_url` to be used.
@penguinoneshaw  let me know what you think about this implementation.

### DIFF
--- a/sdk/__init__.py
+++ b/sdk/__init__.py
@@ -28,13 +28,20 @@ class SDK:
 
     def __init__(
         self,
-        username: str,
-        password: str,
         group_id: str,
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        login_url: Optional[str] = None,
         endpoints: Optional[Endpoints] = None,
         webhook: Optional[str] = None,
     ):
-        self._client = Client(username, password, group_id, endpoints)
+        self._client = Client(
+            username=username,
+            password=password,
+            group_id=group_id,
+            login_url=login_url,
+            endpoints=endpoints,
+        )
         self.batches: Dict[str, Batch] = {}
         self.webhook = webhook
 
--- a/sdk/__init__.py
+++ b/sdk/__init__.py
@@ -16,7 +16,7 @@ import time
 from typing import Any, Dict, List, Optional
 
 from sdk.batch import Batch, RESULT_POLLING_INTERVAL
-from sdk.client import Client
+from sdk.client import Client, TokenProvider
 from sdk.endpoints import Endpoints
 from sdk.job import Job
 from sdk.device.configuration import BaseConfig
@@ -31,7 +31,7 @@ class SDK:
         group_id: str,
         username: Optional[str] = None,
         password: Optional[str] = None,
-        login_url: Optional[str] = None,
+        token_provider: Optional[TokenProvider] = None,
         endpoints: Optional[Endpoints] = None,
         webhook: Optional[str] = None,
     ):
@@ -39,7 +39,7 @@ class SDK:
             username=username,
             password=password,
             group_id=group_id,
-            login_url=login_url,
+            token_provider=token_provider,
             endpoints=endpoints,
         )
         self.batches: Dict[str, Batch] = {}
--- a/sdk/client.py
+++ b/sdk/client.py
@@ -11,42 +11,111 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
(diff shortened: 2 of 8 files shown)


awennersteen:
Changelog needs update

sdk/errors.py
```diff
@@ -39,3 +39,7 @@ def __str__(self) -> str:
             f"Error {self.code}: {self.description}\n"
             f"Details: {json.dumps(self.details, indent=2)}"
         )
+
+
+class CannotLoginError(Exception):
```
MatthieuMoreau0, line 44: `LoginError` seems sufficient for the naming

sdk/client.py
```diff
@@ -82,6 +85,11 @@ def _request(
         data: JSendPayload = rsp.json()
         # If account returns unauthorized we attempt to login and retry request
         if rsp.status_code == 401:
+            #  TODO: Improve the interactive login flow
+            if (self.username == "" and self.password == "") and self.login_url:
```
MatthieuMoreau0, line 89: Instead of empty string should we make username and password optional with default values to None ? This check could be replaced by `if not self.username and not self.password`

If we go this way we could also add an error message to the user if all of login_url, user_name and password are falsy

MatthieuMoreau0 (commented):
Just curious : How does the login url work ?

sdk/client.py
awennersteen: I think empty string is also `False`y

MatthieuMoreau0: That's correct I believe! My main point was around defining default values so that we wouldn't need to do username="" when initializing the SDK with the login_url

awennersteen: exactly, and having type consistency there would be a bit easier for me downstreams

anneclairelh:
@MatthieuMoreau0 @awennersteen I addressed your comment and also after discussion with James did a refactor of the client to use authenticators and token providers. This way for jupyter hub the only thing is to provide a custom token provider like so:

```
class JupyterHubTokenProvider(TokenProvider):
    def __init__(self, hub_api_url: str, hub_api_token: str):
        self.hub_api_url = hub_api_url
        self.hub_api_token = hub_api_token

    def _query_token(self) -> str:
        return requests.get(
            self.hub_api_url, headers={"Authorization": f"Bearer {self.hub_api_token}"}
        ).json()["auth_state"]["access_token"]
```

sdk/client.py
```diff
@@ -59,13 +134,33 @@ def _login(self) -> None:
...
-            "content-type": "application/json",
-            "authorization": f"Bearer {self._token}",
-        }
+
+class Client:
+    authenticator: AuthBase
+
+    def __init__(
+        self,
+        group_id: str,
```
awennersteen, line 145: why is this one not Optional here?

sdk/client.py
```diff
@@ -49,7 +124,7 @@ def _login(self) -> None:
         }
 
         rsp = requests.post(
-            url,
+            self.login_url,
```
MatthieuMoreau0, line 127: We build a URL above but do not use it ?

sdk/__init__.py
```diff
@@ -28,13 +28,20 @@ class SDK:
 
     def __init__(
         self,
-        username: str,
-        password: str,
         group_id: str,
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        login_url: Optional[str] = None,
```
MatthieuMoreau0, line 34: Should this have a convenient default value for the users ? (Is it even necessary anymore ?)

awennersteen: To me I'd still prefer having the default being "" when possible

awennersteen:
Thanks!
Indeed this gives me all that's needed

sdk/client.py
anneclairelh: Sorry, remnant from earlier version.

sdk/__init__.py
anneclairelh: Ha, I've changed everything now... ;) Why so?

awennersteen: It's just an unnecessary type instability (in lack of a better word). For dict, str, list etc it's quite nice (and makes mypy extra happy) to leave them as their empty variants.

On the other hand, it's perhaps more user-instructing to mark them as Optional

awennersteen (approved):
LGTM

PR #67, opened 2023-03-21
Alg/get device specs

Augustinio:
cf changelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,3 +121,15 @@ from sdk.device.configuration import EmuFreeConfig
 configuration = EmuFreeConfig(with_noise=True)
 batch = sdk.create_batch(serialized_sequence, [job1,job2], device_type=DeviceType.EMU_FREE, configuration=configuration)
 ```
+
+### List of supported device specifications
+
+The SDK provides a method to retrieve the device specs supported by Pasqal.
+
+```python
+sdk.get_device_specs_list()
+```
+
+The method returns a list of Pydantic `DeviceSpecs` objects with two parameters: `device_type` and `specs`. The `specs` are
+in the `dict` format. They can be dumped into a json string and deserialized into a `Device` instance from the `Pulser`
+library with the `deserialize_device` method from `Pulser`.
--- a/sdk/device/device_specs.py
+++ b/sdk/device/device_specs.py
@@ -1,0 +1,21 @@
+import json
+from typing import Any, Dict, List
+
+from pydantic import BaseModel, validator
+
+from sdk.device.device_types import DeviceType
+
+
+class DeviceSpecs(BaseModel):
+
+    device_type: DeviceType
+    specs: Dict[str, Any]
+
+    @validator("specs", pre=True)
+    def _convert_specs_to_dict(cls, v):
+        return json.loads(v)
+
+
+class DeviceSpecsList(BaseModel):
+
+    device_specs_list: List[DeviceSpecs]
--- a/tests/fixtures/api/devices/specs/_.GET.json
+++ b/tests/fixtures/api/devices/specs/_.GET.json
@@ -2,6 +2,9 @@
   "code": 200,
   "data": [
     {
+      "id": "00000000-0000-0000-0000-000000000001",
+      "created_at": "2020-01-01T00:00:00+00:00",
+      "updated_at": "2020-01-01T00:00:00+00:00",
       "device_type": "QPU",
       "specs": "{\"version\":\"1\",\"channels\":[],\"name\":\"device\",\"dimensions\":2,\"rydberg_level\":51,\"min_atom_distance\":0.2,\"max_atom_num\":10,\"max_radial_distance\":5,\"interaction_coeff_xy\":null,\"supports_slm_mask\":false,\"max_layout_filling\":0.5,\"reusable_channels\":false,\"pre_calibrated_layouts\":[],\"is_virtual\":false}"
     }
--- a/tests/test_device_specs.py
+++ b/tests/test_device_specs.py
@@ -1,0 +1,19 @@
+from uuid import uuid4
+
+import pytest
+
+from sdk import SDK
+
+
+@pytest.mark.only
+class TestDeviceSpecs:
+    @pytest.fixture(autouse=True)
+    def init_sdk(self, start_mock_request):
+        self.sdk = SDK(
+            username="me@test.com", password="password", group_id=str(uuid4())
+        )
+
+    def test_get_device_specs_success(self):
+        device_specs_list = self.sdk.get_device_specs_list()
+        assert len(device_specs_list) == 1
+        assert type(device_specs_list[0].specs) == dict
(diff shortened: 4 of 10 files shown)


MatthieuMoreau0 (commented):
Some comments on the Readme otherwise LGTM.

Looks like there are some formatting changes (because previous changes were not properly formatted), maybe we should introduce some black check in the CI at some point or as pre-commit

README.md
```diff
@@ -121,3 +121,15 @@ from sdk.device.configuration import EmuFreeConfig
 configuration = EmuFreeConfig(with_noise=True)
 batch = sdk.create_batch(serialized_sequence, [job1,job2], device_type=DeviceType.EMU_FREE, configuration=configuration)
 ```
+
+### List of supported device specifications
+
+The SDK provides a method to retrieve the device specs supported by Pasqal.
```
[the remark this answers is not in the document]
awennersteen: nittity nit:
as currently defined on PASQAL's cloud platform.

Since these specs can (in theory) vary from day to day

sdk/device/device_specs.py
```diff
@@ -0,0 +1,22 @@
...
+
+    @validator("specs", pre=True)
+    def _convert_specs_to_dict(cls, v: str) -> Dict[str, Any]:
+        specs_dict: Dict[str, Any] = json.loads(v)
+        return specs_dict
+
+
+class DeviceSpecsList(BaseModel):
+
+    device_specs_list: List[DeviceSpecs]
```
awennersteen, line 22: I'm unsure about the usage of this, so this might not be relevant, but would it not make sense to be able to get the names of the devices easily?

tests/test_device_specs.py
```diff
@@ -0,0 +1,19 @@
+from uuid import uuid4
+
+import pytest
+
+from sdk import SDK
+
+
+@pytest.mark.only
```
awennersteen, line 8: did you intend to push this?

README.md
```diff
@@ -121,3 +121,15 @@ from sdk.device.configuration import EmuFreeConfig
...
+
+```python
+sdk.get_device_specs_list()
+```
+
+The method returns a list of Pydantic `DeviceSpecs` objects with two parameters: `device_type` and `specs`. The `specs` are
+in the `dict` format. They can be dumped into a json string and deserialized into a `Device` instance from the `Pulser`
+library with the `deserialize_device` method from `Pulser`.
```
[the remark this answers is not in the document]
awennersteen: there is the PulserPasqal module in Pulser. 

I'm not sure where to but the boundary between the packages but perhaps such an example (or even convenience function) belongs in Pulser and not in the cloud-sdk?


Fully agree on pointing to the relevant docs though

tests/fixtures/api/devices/specs/_.GET.json
```diff
@@ -0,0 +1,14 @@
+{
+  "code": 200,
+  "data": [
+    {
+      "id": "00000000-0000-0000-0000-000000000001",
+      "created_at": "2020-01-01T00:00:00+00:00",
+      "updated_at": "2020-01-01T00:00:00+00:00",
+      "device_type": "QPU",
```
MatthieuMoreau0, line 8: This should be FRESNEL I think

README.md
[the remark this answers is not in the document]
MatthieuMoreau0: Yeah for me the boundary is also very unclear as well

sdk/device/device_specs.py
MatthieuMoreau0: In the current implementation, specs are defined for a given device type (not per device). The device_type is included in the DeviceSpecs so the information is available. For convenience we could maybe expose a dictionnary instead of a list though, what do you guys think ?

README.md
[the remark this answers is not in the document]
HGSilveri: As we discussed in person @MatthieuMoreau0 , the sequence validate against the device specs is intrinsic and happens every time the sequence is constructed. If you are coming from a serialized sequence, the deserialization will rebuild that sequence with the given device and during this process all the necessary validation will occur (including the register validation, by the way).

sdk/device/device_specs.py
awennersteen: But I thought part of the interest here was to really have this info at a per device level?

Not knowing/understanding usage entirely I don't have an opinion about what's the better structure

tests/fixtures/api/devices/specs/_.GET.json
Augustinio: Fresnel is not yet a device in our DeviceType enum. QPU however, is.

README.md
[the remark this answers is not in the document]
Augustinio: I thought it could be useful for sdk users to understand that they could use these specs as such to build their device in the pulser lib. I understand the sdk is a different tool but it doesn't really serve a purpose without the pulser library anyway. But I can remove this comment if we want to keep things separated.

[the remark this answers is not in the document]
awennersteen: I think you should leave it as is, and we should add something on the PulserPasqal side instead?

sdk/device/device_specs.py
Augustinio: @awennersteen to summarise:

- device specs are expected to be shared accross device_types (two fresnel devices should have the same specs)
- calibrated layouts should be individual to a device
- we are now, and for the foreseeable future not expecting to have more than one device per device_type (Fresnel)
- we've therefore attached the calibrated layouts to the device_specs as well

In the future we might need to distinguish between devices but they would still share their specs, only the calibrated layouts should vary.

awennersteen: thanks

README.md
```diff
@@ -121,3 +121,16 @@ from sdk.device.configuration import EmuFreeConfig
 configuration = EmuFreeConfig(with_noise=True)
 batch = sdk.create_batch(serialized_sequence, [job1,job2], device_type=DeviceType.EMU_FREE, configuration=configuration)
 ```
+
+### List of supported device specifications
+
+The SDK provides a method to retrieve the devices specs currently defined on PASQAL's cloud platform.
+they define the physical constraints of our QPUs and these constraints enforce some rules on
```
HGSilveri, line 128: nit:
```suggestion
They define the physical constraints of our QPUs and these constraints enforce some rules on
```

HGSilveri (approved):
LGTM. 
Regarding the boundary with `pulser-pasqal`, I would say that the SDK can go so far as to mention the link with `pulser` and `pulser-pasqal`, but not how to use it (so as to keep the implementation details close to home) - basically, what is being done here.

anneclairelh:
Nit title format ! :)

tests/fixtures/api/devices/specs/_.GET.json
Augustinio: Fixed since we don't use the device type enum.

PR #68, opened 2023-03-24
[FEAT] Add cancel methods (#314)

Mildophin:
Cancel methods were added to the package


sdk/endpoints.py
```diff
@@ -21,8 +21,12 @@
...
     from typing_extensions import Final  # type: ignore
 
-CORE_API_URL: Final[str] = "https://apis.pasqal.cloud/core-fast"
-ACCOUNT_API_URL: Final[str] = "https://apis.pasqal.cloud/account"
+CORE_API_URL: Final[str] = "http://localhost:4200"
+ACCOUNT_API_URL: Final[str] = "http://localhost:4201"
+
+# TODO A changer !!!!
+# CORE_API_URL: Final[str] = "https://apis.pasqal.cloud/core-fast"
+# ACCOUNT_API_URL: Final[str] = "https://apis.pasqal.cloud/account"
```
awennersteen, line 29: Please revert :)

tests/test_batch.py
```diff
@@ -1,3 +1,4 @@
+from pprint import pprint
```
awennersteen, line 1: used for debugging?

.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
awennersteen, line 1: I guess these couple of folders we don't need to push?

tests/test_batch.py
Mildophin: yes, totally forgot this import

sdk/endpoints.py
Mildophin: Yes, sure, forgot that changement too

.idea/.gitignore
Mildophin: not needed yes

tests/test_batch.py
```diff
@@ -68,6 +76,46 @@ def test_create_batch_and_fetch_results(self, request_mock):
...
+            == f"{self.sdk._client.endpoints.core}/api/v1/batches/{self.batch.id}/cancel"
+        )
+
+    def test_cancel_batch_success(self, request_mock):
+        self.batch.cancel()
+        assert (
+            self.batch.status == 'CANCELED'
+        )
+
+    def test_add_job_and_get_job(self, request_mock):
```
MatthieuMoreau0, line 99: We already have a test for the add_job method. This test can focus on the get_job part I think!

tests/test_batch.py
```diff
@@ -68,6 +76,46 @@ def test_create_batch_and_fetch_results(self, request_mock):
...
+        )
+        # the name of the dictionary is the id of the job
+        job_just_added_id = list(self.batch.jobs.keys())[-1]
+        job_requested = self.sdk.get_job(job.id)
+        assert (
+            job_requested.id == job_just_added_id
+        )
+
+    def test_cancel_job_endpoint(self, request_mock):
+        self.batch.cancel_job(self.job_id)
```
MatthieuMoreau0, line 112: - Assert the job status
- Don't forget to test both the SDK class cancel_job method and the Job class cancel method (after you add it)

MatthieuMoreau0:
Please update the name of the PR as well, thanks!

sdk/__init__.py
```diff
@@ -123,3 +123,37 @@ def get_batch(self, id: str, fetch_results: bool = False) -> Batch:
...
+        """Cancel the given batch on the PCS
+
+        Args:
+            id: Id of the batch.
+        """
+        batch_rsp = self._client._cancel_batch(id)
+        batch = Batch(**batch_rsp, _client=self._client)
+        return batch
+
+    def get_job(self, id: str) -> Job:
```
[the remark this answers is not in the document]
awennersteen: I see, what about the `wait` argument?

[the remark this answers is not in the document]
MatthieuMoreau0: That's fair, can you look into this @Mildophin

sdk/job.py
```diff
@@ -1,36 +1,46 @@
...
-        - id: Unique identifier for the batch
+        - errors: Error messages that occurred while processing job.
+        - id: Unique identifier for the job.
         - group_id: ID of the group which the users scheduling the job belong to.
-        - status: Status of the job
-        - result(optional): Result of the job.
-        - variables (optional): dictionnary of variables of the job.
-            None if the associated batch is non-parametrized
+        - status: Status of the job. Possible values are: PENDING, RUNNING, DONE, CANCELED, TIMED_OUT, ERROR, PAUSED.
+        - _client: The client that created the job.
```
MatthieuMoreau0, line 22: nit: A Client instance to connect to PCS

MatthieuMoreau0 (commented):
Small comments, otherwise LGTM nice job!

Can we add the `wait` argument as suggest by Aleksander in the `get_job` method ?

sdk/__init__.py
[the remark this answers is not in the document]
MatthieuMoreau0: Looking good

Mildophin:
Rebase and updated, LMK if it looks good to you

tests/conftest.py
```diff
@@ -56,6 +56,7 @@ def start_mock_request(request_mock):
     request_mock.stop()
 
 
+# Used to prevent warning to be displayed during the run of pytest tests
 @pytest.fixture(autouse=True)
```
MatthieuMoreau0, line 60: There is only a specific test that should raise these warnings right ? So not sure we want to enable this for every test

Mildophin: Well normally it should be only the sdk import tests, however even when they are desactivated, pytest still raise a warning. Not sure where it comes from but it's like pytest run the sdk imports during the tests despite having 0 "from sdk.xx" imports.

Mildophin:
Assertion using pytest.warns added, lmk if it is better to you

MatthieuMoreau0 (approved):
LGTM

PR #69, opened 2023-03-27
[FEAT] SDK login directly to Auth0 [#483]

IMCoins:
Task : https://gitlab.pasqal.com/pcs/public-apis/-/issues/483
end 2 end ; https://gitlab.pasqal.com/pcs/cloud-e2e-tests/-/merge_requests/189
It's ready to be merged :+1:


sdk/authentication.py
```diff
@@ -0,0 +1,121 @@
...
+class Auth0TokenProvider(TokenProvider):
+    def __init__(self, username: str, password: str):
+        self.username = username
+        self.password = password
+
+        #   This is public, so we can put it in the code here.
+        self.public_client_id = "5QtfSu1UV118Iz6By6IJRSNoDrLbAiOv"
+        self.audience = "https://apis.dev.pasqal.cloud/account/api/v1"
+        self.realm = "pcs-users"
+        self.domain = "pasqal-dev.eu.auth0.com"
```
penguinoneshaw, line 104: I'm not sure this should be pointing to the dev environment by default.

penguinoneshaw (changes requested):

sdk/authentication.py
IMCoins: You might be right, I didn't even notice the `-dev` tbh. Nice catch

tests/fixtures/auth0/token_response.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
IMCoins: I'm so confused as to why I did that when now you point that detail

sdk/authentication.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+        expires_in = token_response.get("expires_in", None)
+        if expires_in:
+            return datetime.now(tz=timezone.utc) + timedelta(seconds=float(expires_in))
+
+        try:
+            # We assume the token is valid, but might not be in an expected format
+            token = token_response.get("access_token")
+
+            if (isinstance(token, str)and "." in token):
```
penguinoneshaw, line 81: Could you run Black on the changed files, it should fix spacing issues such as this

CHANGELOG.md
```diff
@@ -2,6 +2,19 @@
...
+### Added
+    - A (non-breaking) functionality that allows the user to not pass the password, and secretly prompt it when using the SDK.
+    - Adding local tests to check the login behavior.
+
+### Changed
+    - The authentication now directly connects to the Auth0 platform instead of connecting through PasqalCloud.
+    - Small refactor of files, with the authentication modules in the `authentication.py` file, instead of `client.py`.
+
+### Deleted
+    - Account endpoint, we now use the Auth0.
```
penguinoneshaw, line 16:
```suggestion
    - Account endpoint, we now use Auth0.
```

README.md
```diff
@@ -29,6 +29,52 @@ To run the tutorials or the test suite locally, run the following to install the
...
+```python
+from sdk import SDK, Endpoints
+
+class MyNewEndpoints(Endpoints):
+    ...
+
+sdk = SDK(..., endpoints=MyNewEndpoints)
+```
+
+Doing so will result in you only overriding the endpoints you wish, and be able to use the exising ones.
```
MatthieuMoreau0, line 74: typo: existing

Overall I think this paragraph is not very clear, I'd suggest something along the lines of: "This enables you to target backend services running locally on your machine, or other environment like preprod or dev."

README.md
```diff
@@ -29,6 +29,52 @@ To run the tutorials or the test suite locally, run the following to install the
...
+sdk = SDK(username=username, group_id=group_id)
+> Please, enter your password:
+
+""" Method 3: Use a token.
+    If you already know your token, you can directly pass it as an argument.
+"""
+sdk = SDK(token_provider=token, group_id=group_id)
+```
+
+If you want to redefine the APIs used by the SDK, please, do the following.
```
MatthieuMoreau0, line 63: We don't expect regular users to use this so I think this could be added to a separate section or at least introduced with a comment that says this is intended for development purposes

mypy.ini
```diff
@@ -9,3 +9,4 @@ ignore_errors = false
 warn_redundant_casts = true
 allow_untyped_calls = true
 allow_any_generics = true
+ignore_missing_imports= true
```
MatthieuMoreau0, line 12: Why is this needed ? Could we filter out false positives instead if they are not too common ?

README.md
```diff
@@ -29,6 +29,52 @@ To run the tutorials or the test suite locally, run the following to install the
...
+    If you only want to insert your username, but want a solution to have your password being secret
+    you can run the SDK without password. A prompt will then ask for your password
+"""
+sdk = SDK(username=username, group_id=group_id)
+> Please, enter your password:
+
+""" Method 3: Use a token.
+    If you already know your token, you can directly pass it as an argument.
+"""
+sdk = SDK(token_provider=token, group_id=group_id)
```
MatthieuMoreau0, line 60: It looks like the `token_provider` should be a `TokenProvider` implementation, but you set it above to a string. This is going to raise a `RuntimeError` when calling the `get_token` method, unless I missed something

sdk/authentication.py
```diff
@@ -0,0 +1,123 @@
...
+        return None
+
+
+class Auth0TokenProvider(TokenProvider):
+    def __init__(self, username: str, password: str):
+        self.username = username
+        self.password = password
+
+        #   This is public, so we can put it in the code here.
+        self.public_client_id = "5QtfSu1UV118Iz6By6IJRSNoDrLbAiOv"
```
MatthieuMoreau0, line 102: This depends on the auth0 tenant and thus cannot be hardcoded if we want to make the auth0 endpoint configurable.

I'm not sure sure it would fit well inside the current `Endpoints` class, the naming doesn't sound very adequate.. Maybe it could be fine for now to add it there, i.e. make auth0 a dict  (or dataclass) with a domain key and a client_id key

tests/conftest.py
```diff
@@ -4,34 +4,52 @@
...
+def mock_auth0_response():
+    json_path = "tests/fixtures/auth0/token_response.json"
+    with open(json_path) as json_file:
+        result = json.load(json_file)
+
+    return result
+
+
+def mock_response(request, context):
+    if request.url.startswith(Endpoints.core):
```
MatthieuMoreau0, line 43: I see you are not mocking the calls to account. Does that mean the SDK isn't using it anymore now that we got rid of the login directly ? We can maybe remove it from the `Endpoints` class then!

tests/test_auth.py
```diff
@@ -0,0 +1,29 @@
...
+    def test_module_getpass_success(self, getpass):
+        getpass.return_value = "password"
+        SDK(group_id = "random", username="hi")
+        getpass.assert_called_once()
+
+    @patch("sdk.client.getpass")
+    def test_module_getpass_no_password(self, getpass):
+        getpass.return_value = ""
+        
+        with pytest.raises(Exception):
```
MatthieuMoreau0, line 17: Can we catch a more specific exception here ?

MatthieuMoreau0 (commented):
Small comments, looks good for the most part thanks !

tests/test_auth.py
IMCoins: I can specify a more adequate error indeed

tests/conftest.py
IMCoins: I deleted the Endpoints to Account, but I have modified it because it's used in the `self.audience` of the authentication.
I think we may want to develop a more simple system for the sdk that will change the baseurl based on the `local | dev | preprod | prod` environment. But it wasn't in the task I was assigned :)

sdk/authentication.py
IMCoins: Does the `public_client_id` change ? It's my bad then, can you send me tomorrow the real `public_client_id` of the production ? I don't think I have access to it.

But yeah, if we don't want to hardcode it, I think, as I said in another comment, we should maybe make a system for the SDK to connect seemlessly to either `local | dev | preprod | prod` environment

README.md
IMCoins: You're right, I changed it and applied documentation about an untested behavior. I added this example in order to be thorough, but It looks like I wasn't since I didn't add a test for this existing behavior. I'll do it :)

README.md
```diff
@@ -29,6 +29,52 @@ To run the tutorials or the test suite locally, run the following to install the
...
+```
+
+If you want to redefine the APIs used by the SDK, please, do the following.
+
+```python
+from sdk import SDK, Endpoints
+
+class MyNewEndpoints(Endpoints):
```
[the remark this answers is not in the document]
IMCoins: Yes, but I think it should be recommended to use this feature this way because we don't want to override every endpoints.
Plus, we now have an Endpoint we may not want to change (especially in local mode, since it does not exist) : Auth0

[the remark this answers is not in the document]
awennersteen: But we can just modify specific values in the Endpoints, right?

README.md
```diff
@@ -66,11 +112,9 @@ Once you have serialized your sequence, you can send it with the SDK with the fo
 from sdk import SDK
 from pulser import devices, Register, Sequence
 
-username="your_username" # Replace this value by your username or email on the PASQAL platform.
-group_id="your_group_id" # Replace this value by your group_id on the PASQAL platform.
-password="your_password" # Replace this value by your password on the PASQAL platform.
-# Ideally, do not write this password in a script but provide in through the command-line or as a secret environment variable.
-
+username="your_username" 
+group_id="your_group_id"
```
awennersteen, line 116: I think we should mention that the user has to go to the portal and fetch it from there

mypy.ini
awennersteen: You can override this specifically for the auth0 package then

sdk/client.py
```diff
@@ -147,18 +43,19 @@ def __init__(
...
-        if not (username and password) and not token_provider:
-            raise Exception(
-                "At least a username/password combination or"
-                "TokenProvider object should be provided."
+        if not username and not token_provider:
+            raise ValueError(
+                "At least a username or TokenProvider object should be provided."
             )
+        
+        if password is None and username is not None:
```
awennersteen, line 51: could you make this
```
if username and not password:
```
?
would catch an empty string too, which is more convenient for env var setups

README.md
```diff
@@ -29,6 +29,52 @@ To run the tutorials or the test suite locally, run the following to install the
...
+
+There are several ways to provide a correct authentication using the SDK.
+
+```python
+from sdk import SDK
+
+group_id="your_group_id" # Replace this value by your group_id on the PASQAL platform.
+some_token="token"
+username="your_username" # Replace this value by your username or email on the PASQAL platform.
+password="your_password" # Replace this value by your password on the PASQAL platform.
```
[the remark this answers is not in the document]
IMCoins: Yes, I'll fix this point in my next commit.
But I'm wondering : since we're having this feature, we should document it right ?

README.md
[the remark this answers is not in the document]
IMCoins: Same point as above for the Token, I think since the feature exists and is an entrypoint in the application, we should document it. Yes, for me it is trivial to change the `Endpoints` in the library if I need to make a change using the `pip install -e`, but this feature exists specifically to override the Endpoints, so, I think we should either delete the feature, or document it properly (i.e : give a how to use example)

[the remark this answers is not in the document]
MatthieuMoreau0: Yes, to override the core endpoint and keep the other as defaults you can simply do `endpoints = Endpoints(core=<my-endpoint>)`

README.md
MatthieuMoreau0: Yes we didn't add explicitly the URL to other envs as to not confuse our users but maybe we could add that

sdk/client.py
IMCoins: Line 46-49, just above, the code checks it already:

```python
        if not username and not token_provider:
            raise ValueError(
                "At least a username or TokenProvider object should be provided."
            )
```

README.md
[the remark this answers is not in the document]
IMCoins: I'll do that, thanks for clarifying, I think you're right

README.md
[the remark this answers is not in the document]
IMCoins: So, do you prefer I document it like this ?

```python
new_endpoints = Endpoints(core = "http://new_url:new_port")
```

sdk/client.py
awennersteen: what if I input

```
password = ""
username = "my_username"
```

This would not be caught.

README.md
IMCoins: Indeed, but that would be repeated with the new Authentication section. But I'll add the comment again for the group id

sdk/client.py
IMCoins: I added a test to prove it would be catch, see `test_module_no_user_with_password`

README.md
IMCoins: i readded the comments even if i find them redundant with the above authentication comments

sdk/authentication.py
IMCoins: i changed the public client id to the prod one.
but indeed, we should in another PR make it configurable

mypy.ini
IMCoins: done

README.md
IMCoins: done

sdk/authentication.py
IMCoins: done

README.md
```diff
@@ -29,6 +29,52 @@ To run the tutorials or the test suite locally, run the following to install the
...
+# Ideally, do not write this password in a script but provide in through the command-line or as a secret environment variable.
+
+""" Method 1 : Username + Password
+    If you know your credentials, you can pass them to the SDK instance at instanciation.
```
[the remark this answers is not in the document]
IMCoins: done

tests/fixtures/auth0/token_response.json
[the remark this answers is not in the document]
IMCoins: done

README.md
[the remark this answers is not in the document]
awennersteen: Yeah, that's exactly what I meant too. Sorry for being unspecific!

sdk/client.py
awennersteen: Thanks.  I agree the current implementation will not yield an invalid instance of the SDK.

What I was actually concerned about was this:
```
$ export MY_ENV_VAR=
$ python3
Python 3.10.6 (main, Mar 10 2023, 10:55:28) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> a = os.environ["MY_ENV_VAR"]
>>> a is None
False
>>> type(a)
<class 'str'>
>>> 
```

It might seem artificial to do `export MY_ENV_VAR=`, however, this can easily happen in practice. Especially when the sdk is used as part of an application or a different library.

Which is why, when we're first adding the `getpass` functionality, it would be great for empty password to also be caught in that functionality. If, instead of testing `if password is None` we just test `if password`, this is caught.

Sorry for being vague before, and for laboring on this point, but I think it will save us some time in the future.

sdk/client.py
```diff
@@ -147,20 +43,56 @@ def __init__(
...
+            raise err
+
+    def _credential_login(
+        self, username: str, password: Optional[str]
+    ) -> TokenProvider:
+        if password is None:
+            password = getpass("Enter your password:")
+
+        if not password:
+            raise ValueError("You cannot provide an empty password.")
```
awennersteen, line 92: This is where I meant that it would be nice, in the first `if` statement, to check `if not password`. It would catch someone giving an empty string `""` as the value and let them give the password with the `getpass` functionality.

sdk/client.py
awennersteen: Thanks! Your latest changes look like a great improvement!
I still had one minor nit (arguably also personal preference).

sdk/client.py
awennersteen: hmm, I see your point of view.

However, a lot of the usage of the cloud-sdk is/will be indirect.
So if we first have this feature I think it should be done in a manner which makes it more likely that people can use it.

In much other Pasqal code we avoid `Optional` and default e.g. strings to `""`, so this checking just for `None` will never be reached.

sdk/authentication.py
```diff
@@ -0,0 +1,154 @@
...
+        super().__init__(username, password)
+
+        #   Makes a call in order to check the credentials at creation
+        #   And raise the Error
+        self._query_token()
+
+    def _query_token(self) -> dict[str, Any]:
+        raise Auth0Error(
+            status_code=403, error_code=403, message="Wrong email or password"
+        )
```
MatthieuMoreau0, line 154: I assume this is used for testing purposes ? Can we move this to the tests folder ?

sdk/client.py
```diff
@@ -147,20 +43,58 @@ def __init__(
...
+        err = ValueError("token_provider must be a TokenProvider subclass")
+        try:
+            # The type ignore is because I wouldn't know how to fix the type problem
+            # but the code should be correct, and is tested
+            if not issubclass(token_provider, TokenProvider):  # type: ignore
+                raise err
+        except (
+            TypeError
+        ):  # This is the error if token_provider is not a class for issubclass
+            raise err
```
MatthieuMoreau0, line 83: Why not simply return the TypeError raised by issubclass ?

If we don't like that error message, and want our own `ValueError`, something like below sounds sufficient and simpler:
```python
try:
  issubclass(token_provider, TokenProvider)
except TypeError:
  raise ValueError(<error message>)
```

According to the [doc](https://docs.python.org/3/library/functions.html#issubclass), the `issubclass` function returns True or raises an error so the if statement looks unnecessary to me IIUC

sdk/client.py
```diff
@@ -147,20 +43,58 @@ def __init__(
...
+        self.endpoints = self._make_endpoints(endpoints)
+        self.group_id = group_id
+
+    @staticmethod
+    def _make_endpoints(endpoints: Optional[Endpoints]) -> Endpoints:
+        if endpoints is None:
+            return Endpoints()
+
+        if not isinstance(endpoints, Endpoints):
+            raise ValueError("Endpoints must be a Endpoints instance")
```
MatthieuMoreau0, line 65: TypeError seems more appropriate, no ?

sdk/client.py
```diff
@@ -147,20 +43,58 @@ def __init__(
...
+
+        return endpoints
+
+    @staticmethod
+    def _check_token_provider(token_provider: Optional[TokenProvider]) -> None:
+        if not token_provider:
+            return
```
MatthieuMoreau0, line 72: nit: Personnal preference but I would not allow None token_provider in the `check_token_provider` function since strictly speaking None is not a valid value, instead I would call this function in the init function only if the user inputs a token_provider.

sdk/authentication.py
IMCoins: To me, it's good practice to have those implementations here so that, if we make a modification to the class above, they need to be reflected on those here. :)
But if that sounds like a problem to you, I'll move them

sdk/client.py
IMCoins: i dont see why i shouldn't do that, yeah :+1: i'll do that

sdk/client.py
IMCoins: done

sdk/client.py
IMCoins: done

sdk/client.py
IMCoins: done

sdk/authentication.py
IMCoins: I tried moving them, I had a few issues and I simply didn't want to adress them since we agreed with Matthieu in private that this solution was acceptable as well and both implementations were OK
so, we can say it's ''done''

sdk/authentication.py
```diff
@@ -0,0 +1,146 @@
...
+            pass
+        return None
+
+
+class Auth0TokenProvider(TokenProvider):
+    def __init__(self, username: str, password: str, auth0: Auth0Conf):
+        super().__init__(username, password, auth0)
+
+        # Makes a call in order to check the credentials at creation
+        self._query_token()
```
MatthieuMoreau0, line 107: I understand what you are going for but I feel like this is a bit wasteful since this doesn't update the `token_cache`

sdk/client.py
```diff
@@ -146,21 +41,66 @@ def __init__(
...
+        except TypeError:
+            raise TypeError("token_provider must be a TokenProvider subclass")
+
+    def _credential_login(
+        self, username: str, password: Optional[str], auth0: Auth0Conf
+    ) -> TokenProvider:
+        if not password:
+            password = getpass("Enter your password:")
+
+        #   We want to allow an empty string at first, but then
```
MatthieuMoreau0, line 97: nit: "The prompted password should not be empty"

sdk/client.py
```diff
@@ -146,21 +41,66 @@ def __init__(
...
-                username, password, f"{self.endpoints.account}/api/v1/auth/login"
-            )
+        self.endpoints = self._make_endpoints(endpoints)
+
+        if username:
+            auth0 = self._make_auth0(auth0)
+            token_provider = self._credential_login(username, password, auth0)
 
         self.authenticator = HTTPBearerAuthenticator(token_provider)
+        self.group_id = group_id
```
MatthieuMoreau0, line 60: nit: I feel like this is a bit inconsistent in terms of naming and implementation. 
Some of these are "check", some are "make", there is "credential_login" which builds a TokenProvider and does not login. Could we try to uniformize it a bit and define clearly what is the goal of these functions (user input validation, set defaults, .. ) ?

sdk/client.py
```diff
@@ -146,21 +41,66 @@ def __init__(
         password: Optional[str] = None,
         token_provider: Optional[TokenProvider] = None,
         endpoints: Optional[Endpoints] = None,
+        auth0: Optional[Auth0Conf] = None,
     ):
-        if not (username and password) and not token_provider:
-            raise Exception(
-                "At least a username/password combination or"
-                "TokenProvider object should be provided."
+        if not username and not token_provider:
```
MatthieuMoreau0, line 46: This complex logics warrants a docstring which explains the different possibilities for user to instantiate the Client, with examples. Something like in the Readme imo (maybe we could even omit the token provider part in the readme and add it here only)

tests/test_client.py
```diff
@@ -0,0 +1,129 @@
...
+@patch("sdk.client.Auth0TokenProvider", FakeAuth0BadAuthentication)
+class TestAuthFailure:
+    group_id = "random_group_id"
+    username = "random_username"
+    no_username = ""
+    password = "random_password"
+    no_password = ""
+
+    @patch("sdk.client.getpass")
+    def test_module_getpass_bad_password(self, getpass):
```
MatthieuMoreau0, line 55: This should be moved to the TestAuthInvalidClient no ? Maybe I misunderstand

tests/test_client.py
```diff
@@ -0,0 +1,129 @@
...
+        new_auth0 = Auth0Conf(domain="new_domain")
+        SDK(
+            group_id=self.group_id,
+            username=self.username,
+            password=self.password,
+            auth0=new_auth0,
+        )
+
+
+@patch("sdk.client.Auth0TokenProvider", FakeAuth0BadAuthentication)
```
MatthieuMoreau0, line 46: nit: Can we rename FakeAuth0BadAuthentication to follow the same nomenclature as the test class for consistency ?

sdk/authentication.py
```diff
@@ -0,0 +1,146 @@
...
+
+class FakeAuth0BadAuthentication(TokenProvider):
+    def __init__(self, *args: tuple, **kwags: dict):
+        """The arguments are not important.
+        What's important is that the init raise the error below.
+        """
+        self._query_token()
+
+    def _query_token(self) -> dict[str, Any]:
+        raise ValueError()
```
MatthieuMoreau0, line 146: This should be the same error as the one returned by auth0 when the credentials are incorrect and we should test manually that the resulting error message is clear for the user (and if not catch the Auth0 error and raise a custom one).

IMCoins: This was originally a `Auth0Error`, but I noticed that I didn't want to test for a Auth0Error in Mocked Test, in order to be able to differentiate if the query was mocked or not. If the query is not Mocked, how can I be sure that the error is from the Auth0Mock ? Or the Auth0 real client ?

I saw this because I placed a bad test inside `TestAuthInvalidClient`, but it made a real request, that was green in the test because I was checking for Auth0Error. If I had been checking for `ValueError`, I would have gotten a `Auth0Error` in the test that wasn't mocked (and therefore a red light, what I wanted).

What do you think ?

tests/test_client.py
IMCoins: The `bad_password` is the simulation of a password that is valid, but won't pass the Authentication. Unlike the `no_password`, which will raise an error not even trying to send the request

sdk/client.py
IMCoins: To me, it's crystal clear, but maybe I was too much into the feature myself to realize it's not clear.
The `_make_xxx` is used because we have a parameter, and it's used to assure yourself that the returned value is valid for your implementation.
The `_credential_login` actually makes a login request, and therefore gets a `token` that will be used in the `HTTPBearerAuthenticator`.
The `_check_...` does not return something and only check if the value is correct.

Maybe we want to make a convention in order to uniformize this ?

sdk/authentication.py
IMCoins: What do you propose we do ? I have to change the tests and perform a get_batch() everytime If I want to check that the Auth is working.

tests/test_client.py
IMCoins: done

sdk/client.py
IMCoins: done

sdk/authentication.py
```diff
@@ -0,0 +1,146 @@
...
+            "access_token": "some_token",
+            "id_token": "id_token",
+            "scope": "openid profile email",
+            "expires_in": 86400,
+            "token_type": "Bearer",
+        }
+
+
+class FakeAuth0BadAuthentication(TokenProvider):
+    def __init__(self, *args: tuple, **kwags: dict):
```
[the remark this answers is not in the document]
IMCoins: done

sdk/client.py
IMCoins: I'm confused because the logic is the same as before, but yeah, I'll do it...

IMCoins:
```python
        """ In order to login, you have to provide either a login/password,
            or a token_provider.
            If you do not provide a password in order to fetch a token, that
            will be used to perform actions from this class, you will be prompted
            to enter one.
        """
```

sdk/__init__.py
```diff
@@ -33,14 +34,22 @@ def __init__(
...
         endpoints: Optional[Endpoints] = None,
+        auth0: Optional[Auth0Conf] = None,
         webhook: Optional[str] = None,
     ):
+        """In order to login, you have to provide either a login/password,
+        or a token_provider.
+        If you do not provide a password in order to fetch a token, that
+        will be used to perform actions from this class, you will be prompted
+        to enter one.
+        """
```
MatthieuMoreau0, line 45: suggestion:

"""This class provides helper methods to call the PASQAL Cloud endpoints. 

To authenticate to PASQAL Cloud, you have to provide either an email/password combination or a TokenProvider instance. You may omit the password, you will then be prompted to enter one.

sdk/authentication.py
MatthieuMoreau0: Could we call the `get_token` so that we store it in the cache, maybe ?

sdk/authentication.py
MatthieuMoreau0: You can test directly that your mock was called using `assert_called_once` or equivalent instead of returning a different error. Theoretically speaking I'd rather raise the same errors as the object we are mocking for peace of mind :)

sdk/client.py
MatthieuMoreau0: It's fine

tests/test_client.py
MatthieuMoreau0: Ok thanks for the clarification!

sdk/__init__.py
IMCoins: done

sdk/authentication.py
IMCoins: done

sdk/authentication.py
IMCoins: done

MatthieuMoreau0 (approved):

sdk/authentication.py
```diff
@@ -0,0 +1,146 @@
...
+
+class FakeAuth0AuthenticationFailure(TokenProvider):
+    def __init__(self, *args: tuple, **kwargs: dict):
+        #   Only to raise an error as the Authentication is failed.
+        self._query_token()
+
+    def _query_token(self) -> dict[str, Any]:
+        raise Auth0Error(
+            status_code=403, error_code=403, message="Wrong email/password"
+        )
```
penguinoneshaw, line 146: I'm not altogether happy about this being in the "live" code

penguinoneshaw (changes requested):

CHANGELOG.md
```diff
@@ -2,6 +2,19 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.1.15] 2023-03-27
+
+### Added
+    - Adding tests to check the login behavior.
+    - Adding tests to check the override Endpoints behavior.
```
penguinoneshaw, line 9: Added (inconsistent tense)

penguinoneshaw (approved):

PR #70, opened 2023-03-28
[DOC] Tell users to serialize sequence with abstract representation

MatthieuMoreau0:
The backend expects sequences serialized with the abstract representation. The documentation was outdated on that front.


MatthieuMoreau0:
Similar change has already been applied in https://github.com/pasqal-io/pasqal-cloud/pull/74

Closing

PR #72, opened 2023-04-06
[IMPROVEMENT] Replace DeviceType with EmulatorType

Augustinio:
https://gitlab.pasqal.com/pcs/public-apis/-/issues/527

Cf changelog


Augustinio:
@anneclairelh for some reason I can't reply directly to your comment.

Basically the QPU DeviceType was never really used. It was just a flag to indicate that we weren't using emulators. Since QPU device types are extracted from the sequence.

In public-apis we're removing QPU as a device type since it's not an actual device type and lead to confusion regarding what it meant. Similarly it's a way to make the code clearer and avoid users to potentially get confused with the meaning of this QPU device type.

MatthieuMoreau0:
@anneclairelh 
> Of course, but my question is, do we know that we won't have other QPU device types ? Is the intention behind this that we don't want to expose device types like "FRESNEL" in a public repo ?

We will have other QPU device type, we just don't want to hard code it inside the SDK, as these can be fetched directly from the cloud device specs anyway (and we don't want to have to update the package everytime we add a QPU to the list). Moreover to select a precise QPU, the user will select it directly inside their pulser sequence. 

Having an extra argument on the SDK was making things confusing, the user would select device_type=QPU but the resulting batch would have its attribute device_type set to something like IROISEMVP

README.md
```diff
@@ -162,23 +161,23 @@ For EMU_TN you may add the integrator timestep in nanoseconds, the numerical acc
...
 ```
 
 For EMU_FREE you may add some default SPAM noise. Beware this makes your job take much longer.
 
 ```python
 # replace the corresponding section in the above code example with this to
 # add further configuration
-from sdk.device.device_types import DeviceType
+from sdk.device.emulator_types import EmulatorType
 from sdk.device.configuration import EmuFreeConfig
```
MatthieuMoreau0, line 177: nit: for convenience for the users, could we export these objects in the `sdk.device` module ?

MatthieuMoreau0 (approved):
Nice job LGTM, one tiny suggestion

README.md
Augustinio: Done

MatthieuMoreau0 (approved):

MatthieuMoreau0 (changes requested):
This introduces breaking changes can you bump the minor instead of the patch please to follow semantic versioning ? See #73

MatthieuMoreau0 (approved):

PR #78, opened 2023-05-29
[REFACT] Rename group to project

Mildophin:
Here is the updated version of the sdk I made to make it compatible with either group or project, the goal is open a transition from 'group' to 'project' for the sdk users, without enforcing them to update their code.


pasqal_cloud/__init__.py
```diff
@@ -29,22 +29,32 @@ class SDK:
...
         You may omit the password, you will then be prompted to enter one.
         """
+
+        # Ticket (#622), to be removed, used to avoid a breaking change during the group to project renaming
+        if not (project_id or group_id):
+            raise TypeError("Either a group_id or project_id has to be given as argument")
+
+        # Ticket (#622), to be removed, used to avoid a breaking change during the group to project renaming
+        if not project_id:
+            project_id = group_id
```
Augustinio, line 54: Could we issue a warning regarding the use of a deprecated argument here?

pasqal_cloud/batch.py
```diff
@@ -23,7 +23,7 @@ class Batch:
         - created_at: Timestamp of the creation of the batch.
         - updated_at: Timestamp of the last update of the batch.
         - device_type: Type of device to run the batch on.
-        - group_id: ID of the owner group of the batch.
+        - project_id: ID of the owner project of the batch.
```
Augustinio, line 26: Although `group_id` is deprecated it is still an accepted argument for the `__init__` of batch. Therefore it should be documented as existing but deprecated.

pasqal_cloud/batch.py
```diff
@@ -59,9 +58,20 @@ class Batch:
...
     def __post_init__(self) -> None:
         """Post init method to convert the configuration to a BaseConfig object."""
+
+        # Ticket (#622), used to avoid a breaking change during the group to project renaming
+        if not (self.project_id or self.group_id):
+            raise TypeError("Either a group_id or project_id has to be given as argument")
+
+        # Ticket (#622), to be removed as well
+        if not self.project_id:
+            self.project_id = self.group_id
```
Augustinio, line 73: Same comment as above regarding a deprecated warning.

pasqal_cloud/job.py
```diff
@@ -28,16 +28,25 @@ class Job:
...
+    group_id: Optional[str] = None
+    project_id: Optional[str] = None
+
+    def __post_init__(self) -> None:
+        """Ticket (#622), used to avoid a breaking change during the group to project renaming.
+        Method to be entirely removed"""
+        if not (self.project_id or self.group_id):
+            raise TypeError("Either a group_id or project_id has to be given as argument")
+        if not self.project_id:
+            self.project_id = self.group_id
```
Augustinio, line 49: same comment as for batch

pasqal_cloud/client.py
```diff
@@ -36,13 +36,22 @@ class Client:
...
+        group_id: Optional[str] = None,
+        project_id: Optional[str] = None
     ):
+        # Ticket (#622), used to avoid a breaking change during the group to project renaming
+        if not (project_id or group_id):
+            raise TypeError("Either a group_id or project_id has to be given as argument")
+
+        # Ticket (#622),to be removed as well
+        if not project_id:
+            project_id = group_id
```
Augustinio, line 53: The client should be called directly by users of the sdk. Therefore I think it would be ok to only allow `project_id` here and remove these checks/conversion

pasqal_cloud/job.py
```diff
@@ -17,7 +17,7 @@ class Job:
         - batch_id: ID of the batch which the job belongs to.
         - errors: Error messages that occurred while processing job.
         - id: Unique identifier for the job.
-        - group_id: ID of the group which the users scheduling the job belong to.
+        - project_id: ID of the project which the users scheduling the job belong to.
```
Augustinio, line 20: same comment as for batch

Augustinio (changes requested):
Couple minor things to fix. Otherwise looks good.

Mildophin:
Updated following your comments @Augustinio , I agree on the client part, group_id is not really useful.

Augustinio (approved):

Augustinio:
Approved at the condition that e2e tests succeed

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.2.6] - 2023-05-29
+
+### Changed
+
+- Accepts either group_id or project_id for the sdk, batch or job. This is a patch preparing the depreciation of group_id.
```
MatthieuMoreau0, line 9: The changelog is intended for the user, I would phrase it differently, something like:

`"Groups" have been renamed as "Projects", hence URL endpoints and attribute names have been changed accordingly. For example the `group_id` of a batch has been renamed to `project_id`. Note that for backwards compatibility, the `group_id` is still exposed by the APIs as a duplicate of the `project_id`. 

However it is considered deprecated and will be removed in future releases.

pasqal_cloud/batch.py
```diff
@@ -59,9 +60,25 @@ class Batch:
     jobs_count: int = 0
     jobs_count_per_status: Dict[str, int] = field(default_factory=dict)
     configuration: Optional[Union[BaseConfig, dict]] = None
+    # Ticket (#622)
+    group_id: Optional[str] = None
+    project_id: Optional[str] = None
 
     def __post_init__(self) -> None:
```
MatthieuMoreau0, line 67: IIRC this method is intended to parse the response from the API, so here we know that the API will return the project_id (and for now the group_id as well) so I don't think the Exception is required. Maybe I'm missing something though

pasqal_cloud/client.py
```diff
@@ -36,12 +36,12 @@ class Client:
 
     def __init__(
         self,
-        group_id: str,
+        project_id: Optional[str] = None,
```
MatthieuMoreau0, line 39: Why optional ?

pasqal_cloud/job.py
```diff
@@ -28,16 +30,30 @@ class Job:
...
     errors: List[str]
     start_timestamp: Optional[str] = None
     end_timestamp: Optional[str] = None
     result: Optional[Dict[str, Any]] = None
     variables: Optional[Dict[str, Any]] = None
+    # Ticket (#622)
+    group_id: Optional[str] = None
+    project_id: Optional[str] = None
+
+    def __post_init__(self) -> None:
```
MatthieuMoreau0, line 46: same remark

tests/test_project_renaming_compatibility.py
```diff
@@ -0,0 +1,56 @@
...
+    group_id = "random_group_id"
+    project_id = "random_project_id"
+    username = "random_username"
+    password = "random_password"
+
+    def test_client_group_argument_converted_to_project(self):
+        sdk = SDK(group_id=self.group_id, username=self.username, password=self.password)
+        assert sdk._client.project_id == self.group_id
+
+    def test_client_project_and_group_given(self):
```
MatthieuMoreau0, line 26: Looks like this test is not doing what the name implies

pasqal_cloud/client.py
Mildophin: Error on my side for this one, forgot to change it, mb

tests/test_project_renaming_compatibility.py
Mildophin: changed

pasqal_cloud/client.py
Mildophin: Actually i remember now why I used, it is cause of Mypy, because this project_id or group_id of the sdk are optional so the project_id has to be optional too

MatthieuMoreau0: It's optional in the SDK but we raise an error if not set, so mypy should not raise an error if this is strictly set to `str`. However you need to also unset the default value `= None` in the `__init__` function here

pasqal_cloud/client.py
```diff
@@ -36,7 +36,8 @@ class Client:
 
     def __init__(
         self,
-        project_id,
+        # Ticket (#622), this has to don't be optional anymore
```
MatthieuMoreau0: Fix phrasing

Mildophin: I changed it in the last commit, was an amend though

pasqal_cloud/client.py
Mildophin: Even if I changed it I'm still getting error: Argument "project_id" to "Client" has incompatible type "Optional[str]"; expected "str"  [arg-type]

Mildophin:
@MatthieuMoreau0 let me know if it looks good to you

MatthieuMoreau0 (approved):

PR #80, opened 2023-06-05
[FEAT] Create an "env" abstraction layer

Mildophin:


pasqal_cloud/__init__.py
```diff
@@ -49,13 +50,26 @@ def __init__(
...
+                raise TypeError("Either a 'group_id' or 'project_id' has to be given as argument")
+            warn("The parameter 'group_id' is deprecated, from now use 'project_id' instead",
                  DeprecationWarning,
                  stacklevel=2
                  )
             project_id = group_id
 
+        # TBR Ticket slot
+        if endpoints or auth0:
+            warn("The parameters 'endpoints' and 'auth0' are deprecated, from now use 'env' instead",
```
oliver-gordon, line 62: nit: `from now` -> `from now on`

pasqal_cloud/__init__.py
```diff
@@ -49,13 +50,26 @@ def __init__(
...
+                 DeprecationWarning,
+                 stacklevel=2
+                 )
+
+        if env:
+            # TBR Ticket slot
+            if endpoints or auth0:
+                raise TypeError("Either 'endpoints' with 'auth0' OR env parameters must be used, not all together")
```
oliver-gordon, line 70: The message makes it hard to understand the expected logic, can you reword slightly? I'm not actually sure what the limitation is from this alone

pasqal_cloud/endpoints.py
```diff
@@ -16,29 +16,93 @@
 
 from dataclasses import dataclass
 
+import typing
```
oliver-gordon, line 19: Can you import all types up here so we don't have `typing.value` in the code base

pasqal_cloud/endpoints.py
```diff
@@ -16,29 +16,93 @@
...
+
+
+# ---- Environment ----
+
+ENV_OPTIONS = ("prod", "preprod", "dev")
+
+
+def init_endpoints(env: str) -> typing.Tuple[Endpoints, Auth0Conf]:
+    if env not in ENV_OPTIONS:
+        raise ValueError("This environment does not exist, valid options are {}".format(', '.join(ENV_OPTIONS)))
```
oliver-gordon, line 71: Can you use f-strings instead of format? I don't think we're locked on a version of python that doesn't support right?

tests/test_client.py
```diff
@@ -58,6 +66,48 @@ def test_correct_new_auth0(self):
...
+                password=self.password,
+                env="wrong"
+            )
+
+    @pytest.mark.filterwarnings(
+        "ignore:The parameters 'endpoints' and 'auth0' are deprecated, from now use 'env' instead"
+    )
+    def test_endpoints_auth0_env_given(self):
+        with pytest.raises(TypeError,
+                           match="Either 'endpoints' with 'auth0' OR env parameters must be used, not all together"):
```
oliver-gordon, line 100: Is this black formatted? Looks weird

pasqal_cloud/__init__.py
Mildophin: Basically it's to make sure that the user will understand the endpoints the sdk will use as if he provides 'endpoints', 'auth0' and 'env' he won't know which one will be used. This is here to prevent confusion

tests/test_client.py
Mildophin: Was from my initiative, I agree, I will try formatting it differently

pasqal_cloud/__init__.py
```diff
@@ -44,18 +45,34 @@ def __init__(
         To authenticate to PASQAL Cloud, you have to provide either an
         email/password combination or a TokenProvider instance.
         You may omit the password, you will then be prompted to enter one.
+
+        Also, you can use the parameter `env` to have the endpoints and auth0 endpoints filled automatically.
```
MatthieuMoreau0, line 49: This was not done before but since the number of args is starting to pile up, it could be a good opportunity to describe all of them in the docstring and flag the deprecated ones.  

e.g. something like
```python
"""
Function Description

Args:
  username: email of the user to login as
  password: [...]
  [...]
  auth0 (deprecated): Auth0Config object to define the auth0 tenant to target
""" 
```

Maybe put the deprecated arguments last ?

On a side note, I think this sentence here is not very clear. I would say something like "If you want to target a non-prod environment, set the `env` kwarg to the name of the environment to target. This will define automatically for you the auth0 config and the pasqal cloud endpoints.

pasqal_cloud/__init__.py
```diff
@@ -44,18 +45,34 @@ def __init__(
...
+        if endpoints or auth0:
+            warn("The parameters 'endpoints' and 'auth0' are deprecated, from now use 'env' instead",
+                 DeprecationWarning,
+                 stacklevel=2
+                 )
+
+        if env:
+            # TBR Ticket slot
+            if endpoints or auth0:
+                raise TypeError("Parameters 'endpoints' and 'auth0' must be used OR 'env' alone otherwise, not all "
```
MatthieuMoreau0, line 72: Suggestion: The `env` argument is incompatible with `endpoints` and `auth0` arguments.  You may either target a pre-defined environment, or define custom auth0 configuration and API endpoints.

pasqal_cloud/endpoints.py
```diff
@@ -15,30 +15,95 @@
...
+    audience: str = PROD_AUDIENCE
     realm: str = REALM
+
+
+# ---- Environment ----
+
+ENV_OPTIONS = ("prod", "preprod", "dev")
+
+
+def select_env_endpoints(env: str) -> Tuple[Endpoints, Auth0Conf]:
```
MatthieuMoreau0, line 70: How about simply `select_env` ?

pasqal_cloud/endpoints.py
```diff
@@ -15,30 +15,95 @@
...
-    audience: str = AUDIENCE
+    domain: str = PROD_AUTH0_DOMAIN
+    public_client_id: str = PROD_PUBLIC_CLIENT_ID
+    audience: str = PROD_AUDIENCE
     realm: str = REALM
+
+
+# ---- Environment ----
+
+ENV_OPTIONS = ("prod", "preprod", "dev")
```
MatthieuMoreau0, line 67: nit: This could be an enum, and its values reused to define the keys of the `endpoints` dictionnary in the function below

pasqal_cloud/endpoints.py
```diff
@@ -15,30 +15,95 @@
...
+
+
+def select_env_endpoints(env: str) -> Tuple[Endpoints, Auth0Conf]:
+    if env not in ENV_OPTIONS:
+        raise ValueError(f"This environment does not exist, valid options are {', '.join(ENV_OPTIONS)}")
+
+    endpoints = {
```
MatthieuMoreau0, line 74: suggestion: naming this `env` ?

tests/test_client.py
```diff
@@ -58,6 +66,50 @@ def test_correct_new_auth0(self):
             auth0=new_auth0,
         )
 
+    @pytest.mark.parametrize(
+        "env, core_endpoint_expected",
+        [
+            ('prod', 'https://apis.pasqal.cloud/core-fast'),
+            ('preprod', 'https://apis.preprod.pasqal.cloud/core-fast'),
+            ('dev', 'https://apis.dev.pasqal.cloud/core-fast')
```
MatthieuMoreau0, line 74: You should test the None case as well

tests/test_client.py
```diff
@@ -58,6 +66,50 @@ def test_correct_new_auth0(self):
...
+    def test_good_env(self, env, core_endpoint_expected):
+        sdk = SDK(
+            project_id=self.project_id,
+            username=self.username,
+            password=self.password,
+            env=env
+        )
+        assert sdk._client.endpoints.core == core_endpoint_expected
+
+    def test_bad_env(self):
```
MatthieuMoreau0, line 86: `test_init_sdk_with_unknown_env`

On a side note this seems to be in the `TestAuthSuccess` class which doesn't seem appropriate. We may want to either rename the test class or move this out of it

tests/test_client.py
```diff
@@ -58,6 +66,50 @@ def test_correct_new_auth0(self):
...
+                project_id=self.project_id,
+                username=self.username,
+                password=self.password,
+                env="wrong"
+            )
+
+    @pytest.mark.filterwarnings(
+        "ignore:The parameters 'endpoints' and 'auth0' are deprecated, from now use 'env' instead"
+    )
+    def test_endpoints_auth0_env_given(self):
```
MatthieuMoreau0, line 98: Can you add several test: case where user inputs:
- env + auth0 
- env + endpoints
- env + auth0 + endpoints

pasqal_cloud/endpoints.py
Mildophin: the dictionary right ?

pasqal_cloud/__init__.py
Mildophin: I was thinking about doing it as Batch and Job got big explanations, I'm planning to add it

pasqal_cloud/endpoints.py
Mildophin: what about the parameter name ? Feel free to give me other options

tests/test_client.py
```diff
@@ -143,3 +146,84 @@ def test_bad_endpoints(self):
...
+                username=self.username,
+                password=self.password,
+                endpoints=Endpoints(core=self.new_core_endpoint),
+                env="prod"
+            )
+
+    @pytest.mark.filterwarnings(
+        "ignore:The parameters 'endpoints' and 'auth0' are deprecated, from now use 'env' instead"
+    )
+    def test_auth0_and_env_given(self):
```
MatthieuMoreau0, line 217: Can you merge the 3 test above using pytest.mark.parametrize ?

README.md
```diff
@@ -161,23 +161,20 @@ The method returns a dict object mapping a device type to a serialized device sp
 to instantiate a `Device` instance in the `Pulser` library.
 
 
-### Target different API endpoints
+### Using `env` parameter to target different API endpoints
 
 This is intended to the package developers.
```
MatthieuMoreau0, line 166: This is intended to the package developers or users which were given access to non-prod environments of the PASQAL cloud platform.

tests/test_client.py
Mildophin: Working properly now :)

Mildophin:
@MatthieuMoreau0 Updated, I took the option to remove the old content from the readme about endpoints as the users won't tweak it as much as we do, so I guess they only need the dicts I made to make their code working.

tests/test_client.py
```diff
@@ -143,3 +153,42 @@ def test_bad_endpoints(self):
...
+        sdk = SDK(
+            project_id=self.project_id,
+            username=self.username,
+            password=self.password,
+            auth0=AUTH0_CONFIG[env],
+            endpoints=PASQAL_ENDPOINTS[env],
+        )
+        assert sdk._client.endpoints.core == core_endpoint_expected
+
+    def test_init_sdk_with_unknown_auth0_dict_key(self):
```
MatthieuMoreau0, line 178: I respect the effort but this is just testing that python raises a KeyError when trying to access a key that doesn't exist in a dictionary, I don't think we should test that :wink:

tests/test_client.py
```diff
@@ -143,3 +153,42 @@ def test_bad_endpoints(self):
...
+        with pytest.raises(KeyError):
+            SDK(
+                project_id=self.project_id,
+                username=self.username,
+                password=self.password,
+                auth0=AUTH0_CONFIG["wrong"],
+            )
+
+    def test_init_sdk_with_unknown_endpoints_dict_key(self):
```
MatthieuMoreau0, line 187: same remark here

tests/test_client.py
```diff
@@ -143,3 +153,42 @@ def test_bad_endpoints(self):
...
+class TestSDKAuth0AndEndpointsDicts(TestSDKCommonAttributes):
+    @pytest.mark.parametrize(
+        "env, core_endpoint_expected",
+        [
+            ("prod", "https://apis.pasqal.cloud/core-fast"),
+            ("preprod", "https://apis.preprod.pasqal.cloud/core-fast"),
+            ("dev", "https://apis.dev.pasqal.cloud/core-fast"),
+        ],
+    )
+    def test_existing_dict_keys(self, env, core_endpoint_expected):
```
MatthieuMoreau0, line 168: naming suggestion: `test_select_env` 

The name of the test (and the test itself) should be focused on the feature it tests instead of implementation details

MatthieuMoreau0 (commented):
Overall looking nice and easy! Just a few minor comments and should be ready to merge after the e2e tests pass!

There is no version and changelog update is that wanted ?

MatthieuMoreau0 (approved):

PR #81, opened 2023-06-06
[FEAT] Successful instantiation of Batch and Job models with unknown fields

MatthieuMoreau0:
With previous implementation, any time a field was added to the API response to the batches and jobs endpoint, an update to the SDK was required to add the field to the dataclasses; and all previous versions of the SDK became obsolete. Our users do not want (or cannot) update the SDK as soon as a new version is released; so we need our code to allow for these extra fields.

This MR:
- Replaces Job and Batch dataclasses with pydantic models which are more flexible
- Updates some incoherent type hints in Batch and Job classes which made test fail with pydantic
- Moves job dictionary building logic out of client and into the Batch model. To be fair building this dictionnary was a poor design decision from the start but we have to keep it for backwards compatibility.
- Adds tests to make sure Job and Batch can be instantiated even when extra unknown fields are returned by the API 
- Updates black version because it didn't seem to work properly with previously used version
- Applies black formatting to the whole project. For review, I suggest reading commit by commits to not be polluted by formatting changes which were confined to a single commit  

Note there are other dataclasses in our codebase but they are not used to parse API response so I didn't bother migrating them over to pydantic. This can be done later

TODO:
- run e2e tests
- Update changelog
- Bump version


pasqal_cloud/batch.py
```diff
@@ -51,30 +54,43 @@ class Batch:
...
     configuration: Optional[Union[BaseConfig, dict]] = None
-    # Ticket (#622)
-    group_id: Optional[str] = None
 
-    def __post_init__(self) -> None:
-        """Post init method to convert the configuration to a BaseConfig object."""
-        if not isinstance(self.configuration, dict):
-            return
+    class Config:
+        extra = Extra.allow
```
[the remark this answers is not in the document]
Yoric: I believe that it's the right way to do things. Does pydantic store the extra fields inline or in a separate object/dictionary?

pasqal_cloud/batch.py
```diff
@@ -51,30 +54,45 @@ class Batch:
...
     status: str
-    webhook: str
+    webhook: Optional[str]
     _client: Client
     sequence_builder: str
-    start_datetime: Optional[str] = None
-    end_datetime: Optional[str] = None
+    start_datetime: Optional[str]
+    end_datetime: Optional[str]
     device_status: Optional[str] = None
```
Yoric, line 62: Nit: If you're removing all these `= None`, let's remove that one, too :)

Yoric (approved):
LGTM!

pasqal_cloud/batch.py
awennersteen: I only have one doubt here
`Optional[str]` is often expected to be handled automagically by inserting `None` if a value wasn't supplied. This is not how Python behaves though.  You'll still have to supply a value :P

Personally I've always preferred having these as `sart_datetime: str = ""`, since empty string in python evaluates as `False`

pasqal_cloud/authentication.py
```diff
@@ -44,6 +44,7 @@ class ExpiringTokenProvider(TokenProvider, ABC):
     this class and defining your own `_query_token` method. See Auth0TokenProvider for
     an implementation example.
     """
+
```
awennersteen, line 47: should we add a linting stage to the project to check that things have been formatted correctly?

In other PASQAL project we use `pre-commit` hooks a lot, checking black, and flake8 at the very least.

pasqal_cloud/batch.py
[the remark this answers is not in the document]
awennersteen: agree

pasqal_cloud/device/configuration/base_config.py
```diff
@@ -26,7 +26,7 @@ class BaseConfig:
     parameters are added to the configuration dictionary.
     """
 
-    extra_config: Optional[dict[str, Any]] = None
```
awennersteen, line 29: any reason for this? I'm not sure what convention we follow, but the most recent standard says to favour `dict` as the type over `Dict`.

tests/test_batch.py
```diff
@@ -205,3 +201,19 @@ def test_create_batch_configuration(self, emulator, configuration, expected):
             configuration=configuration,
         )
         assert batch.configuration == expected
+
+    def test_batch_instanciation_with_extra_field(self, batch):
```
awennersteen, line 205: Nit: `instantiation`

tests/test_job.py
```diff
@@ -0,0 +1,19 @@
+from pasqal_cloud import Job
+
+
+class TestJob:
+    def test_job_instanciation_with_extra_field(self, job):
```
awennersteen, line 5: nit: `instantiation`

Mildophin:
LGTM after changing the things @awennersteen mentioned. A lot of formatting in this MR, half of it formatted was written by me so I will try paying more attention to that for the next MR I will do on the sdk.

MatthieuMoreau0:
@Mildophin You shouldn't have to worry about formatting, your IDE should take care of it for you. I think the black version we were using was not working properly, I updated the dependency. If you update your venv and you have enabled automatic formatting on save in your IDE it should work. Let me know if you need help setting that up

pasqal_cloud/batch.py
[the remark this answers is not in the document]
MatthieuMoreau0: Yes the extra fields are stored inline as you can see in the unit test that was added

pasqal_cloud/authentication.py
MatthieuMoreau0: I added a black check in the github actions. The `pre-commit` hook can be added in another MR, I created the following issue https://github.com/pasqal-io/pasqal-cloud/issues/82

pasqal_cloud/device/configuration/base_config.py
MatthieuMoreau0: Not using the `Dict` type hint in a pydantic model for python version <3.8 raises errors. See for example CI https://github.com/pasqal-io/pasqal-cloud/pull/81/checks?check_run_id=14051565461

Rule of thumb I am using is to use new type hints (e.g. `dict` and `list`) everywhere except for attributes of a pydantic model

pasqal_cloud/authentication.py
awennersteen: thanks!

pasqal_cloud/device/configuration/base_config.py
awennersteen: I see. Pulser has recently dropped support for 3.7, and it reaches of life in less than 3 weeks (27 Jun 2023).

I'll open an issue for that

pasqal_cloud/batch.py
awennersteen: sounds fine to me to leave it has you have then

PR #87, opened 2023-06-19
[FIX] User ID type hint: int -> str

MatthieuMoreau0:
User IDs are strings since we have moved our authentication system to auth0. This bug wasn't caught until now because:
- we were using dataclasses until very recently which were not applying validation when unserializing data
- after switching to pydantic, this is not caught in our e2e tests because all the users used for the e2e tests are users that existed before the migration to auth0 and hence their IDs were reused when we moved to auth0. Hence they still can be parsed as integers.


pasqal_cloud/batch.py
```diff
@@ -51,7 +51,7 @@ class Batch(BaseModel):
     device_type: str
     project_id: str
     id: str
-    user_id: int
+    user_id: str
```
oliver-gordon, line 54: This the only user_id field to maybe be mistyped? I looked briefly but didn't see

oliver-gordon (approved):
Looks ok

pasqal_cloud/batch.py
MatthieuMoreau0: Yep looked like it

PR #88, opened 2023-06-20
Drop python 3.7

### LINKED ISSUE #83: drop support for Python 3.7
No longer supported by Pulser and it goes eol on June 27th.

With that we can also move to using the new lower-case typehints in Pydantic models
See comment:
https://github.com/pasqal-io/pasqal-cloud/pull/81#discussion_r1221150668

awennersteen:
Closes #83


setup.py
```diff
@@ -28,7 +28,7 @@
     long_description_content_type="text/markdown",
     maintainer="Pasqal Cloud Services",
     maintainer_email="pcs@pasqal.io",
-    python_requires=">=3.7",
+    python_requires=">=3.8",
     license="Apache 2.0",
     # TODO:
     classifiers=[
```
MatthieuMoreau0, line 34: below, can you rm 
```python
        ":python_version == '3.7'": [
            "typing-extensions==4.4.0",
        ],
```

awennersteen:
I just realized that we talked about then using only lowercase types. 
Do we prefer to do that here or in a separate PR?

Issue #66, opened 2023-03-02
If fetch_results=True in sdk.get_batch() we should wait for job completion 


awennersteen:
This would make the behavior consistent with the job submission fetch_results behaviour.

MatthieuMoreau0:
Is this not the case ? That sounds like a bug

awennersteen:
I would say no (but I've not ran and confirmed yet)

As you see in the code (client), there is no check that the batch is done
```
    def _get_batch(
        self, id: str, fetch_results: bool = False
    ) -> Tuple[Dict[str, Any], List[Dict[str, Any]]]:
        batch_data: Dict[str, Any] = self._request(
            "GET", f"{self.endpoints.core}/api/v1/batches/{id}"
        )["data"]
        jobs_data = batch_data.pop("jobs", [])
        if fetch_results:
            results = self._request(
                "GET", f"{self.endpoints.core}/api/v1/batches/{id}/results"
            )["data"]
            for job_data in jobs_data:
                job_data["result"] = results.get(str(job_data["id"]), None)
        return batch_data, jobs_data
```

MatthieuMoreau0:
Ah yes you're right I was thinking of the _create_batch function! Would be nice! I think we could just move all the logic that is inside the `create_batch` function to the `get_batch` function

Do you need this for QD module 2 release  ?

awennersteen:
For now I just left this in a different library (because I wanted to do some other checks too, so it was nice), but as soon as this is released this would be something to fix. But we should probably take the time to consider all the functions and make sure that everything works as expected / desired.

ArnaudPel:
Would also be convenient to support an optional `timeout` argument, for example in test environments where we know that the batch is supposed to take a few seconds at most.

Infinite waiting is a rather annoying experience when running a test suite.

awennersteen:
deprecated fetch_results

PR #89, opened 2023-06-20
[FEAT] Add full result, remove `fetch_result` kwarg.

Augustinio:
See changelog.

It appears results were included in the payload of the batch.jobs all along so the results endpoint is currently redundant and we can cut down on the amount of calls performed.


pasqal_cloud/__init__.py
```diff
@@ -131,35 +128,30 @@ def create_batch(
...
             while batch_rsp["status"] in ["PENDING", "RUNNING"]:
                 time.sleep(RESULT_POLLING_INTERVAL)
-                batch_rsp, jobs_rsp = self._client._get_batch(batch_id)
+                batch_rsp = self._client._get_batch(batch_id)
 
-            if fetch_results:
-                batch_rsp, jobs_rsp = self._client._get_batch(
-                    batch_id, fetch_results=True
-                )
-        batch = Batch(**batch_rsp, jobs=jobs_rsp, _client=self._client)
```
awennersteen, line 145: I'm now a bit confused. 
I'm pretty sure I added this code, in the last 6 months(?), specifically because this didn't happen

awennersteen: Also note:
https://github.com/pasqal-io/pasqal-cloud/issues/66

pasqal_cloud/batch.py
```diff
@@ -123,32 +123,25 @@ def add_job(
...
         will be executed before the batch is terminated. When all its jobs are done,
         the complete batch is unassigned to its running device.
         """
         batch_rsp = self._client._complete_batch(self.id)
         self.complete = True
-        if wait or fetch_results:
+        if wait:
             while batch_rsp["status"] in ["PENDING", "RUNNING"]:
                 time.sleep(RESULT_POLLING_INTERVAL)
-                batch_rsp, jobs_rsp = self._client._get_batch(
```
awennersteen, line 144: this always confused me. But never went and double checked

CHANGELOG.md
```diff
@@ -2,6 +2,19 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.2.9] - 2023-06-20
```
awennersteen, line 5: since you're removing an argument, this is a very breaking change for users.
Could we release v 0.3.0?

CHANGELOG.md
```diff
@@ -18,7 +31,7 @@ Now you can get the environment configuration for `endpoints` and `auth0` of the
 ### Changed
 
 - Batch and Job dataclasses have been replaced by pydantic models, which gives more control to unserialize the API response data.
-The SDK is now more resilient to API changes, if a new field is added to the response of the API then the job and batch object instantiation will not raise an exception, meaning this SK version will not become obsolete as soon as the API spec is updated.
+  The SDK is now more resilient to API changes, if a new field is added to the response of the API then the job and batch object instantiation will not raise an exception, meaning this SK version will not become obsolete as soon as the API spec is updated.
```
awennersteen, line 34: >meaning this SK version
should be SDK?

Mildophin (changes requested):
Not much more to say than what aleksander said. I agree on the breaking change part 0.3.0 would be good to implement I guess.

CHANGELOG.md
Augustinio: Can do

pasqal_cloud/__init__.py
Augustinio: Tested it in dev and it does happen. Wait already checks that the status is complete. Once it's complete the data from the jobs, including results is already part of the batch payload.

Augustinio: Of course we'll wait for the e2e pipeline to run successfully.

Augustinio: Looking at the dgx_connector code, results are sent prior to updating batch status, so I don't see in what scenario a fetched completed batch (which already contains the results) would not have the most up to date results.

awennersteen: Could you also do a manual safety check and ensure this doesn't result in a race condition? 

Also, set the Polling interval to something small, send say a few hundred batches and just verify that nothing goes wrong.
I'm just a bit skeptical because I remember when this code was introduced, and it was after we moved to FastAPI.

It could be that it was some faulty logic in core that has later been fixed. This behavior that this MR expects is of course what is the most sensible

Augustinio: I'll run the check with a couple (hundred?) batches in parallel.
Maybe it was before we fixed some issues with async sqlalchemy sessions.
But  even then since the order of operations is as follows:
- send result for all jobs
- check this is the last job
- set batch status to complete
I'm unsure how a complete batch could have non up-to-date results

pasqal_cloud/__init__.py
```diff
@@ -131,35 +128,30 @@ def create_batch(
...
+        batch = Batch(**batch_rsp, _client=self._client)
```
MatthieuMoreau0, line 138: Are the jobs also returned by the get batch endpoint ?

MatthieuMoreau0 (commented):
Looking good, but indeed the previous implementation is quite confusing. I assume we changed the behaviour of the API and didn't adjust the SDK accordingly

pasqal_cloud/__init__.py
Augustinio: yes

pasqal_cloud/__init__.py
Augustinio: @awennersteen following our conversation, if you still have doubts regarding the viability of this approach, would you mind pulling from this branch and running your usual tests battery? 
Otherwise can you resolve?

Mildophin (approved):

CHANGELOG.md
MatthieuMoreau0: Should we keep the argument and deprecate it so that we don't have to bump the minor ?

pasqal_cloud/__init__.py
awennersteen: it's fine for me

CHANGELOG.md
awennersteen: This would be the smoothest transition.
Keep them in the `create_batch` and `get_batch` functions, mark as deprecated and ignore.

Augustinio: OK

awennersteen (approved):
LGTM

PR #90, opened 2023-06-22
[FEAT] add `result_type` to config

Augustinio:
Cf Changelog.

At the moment this feature doesn't provide anything new. Users can select a result_type but it's limited to one: `counter` which is already the default for the current devices in `public_apis`. For that reason I did not document the feature yet.


MatthieuMoreau0 (approved):
lgtm, don't forget to bump version

awennersteen (approved):
lgtm

PR #93, opened 2023-07-04
[FIX] Reinstate fetch results

anneclairelh:
Removing the functionality of this argument was not required _and_ broke some user scripts by breaking the integration with `qoolbox`. This reinstates the functionality of the argument while keeping the DeprecationWarning.

PR #94, opened 2023-07-05
[FIX] Reinstate fetch results

anneclairelh:
Removing the functionality of this argument was not required and broke some user scripts by breaking the integration with qoolbox. This reinstates the functionality of the argument while keeping the DeprecationWarning.

(Moved over from https://github.com/pasqal-io/pasqal-cloud/pull/93 bc of a Github issue).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,4 +1,12 @@
+[tool.black]
+line-length = 88
+
+[tool.isort]
+profile = "black"
+skip_gitignore = true
+order_by_type = false
+force_alphabetical_sort_within_sections = true
+
 [build-system]
 requires = ["setuptools>=42", "wheel"]
 build-backend = "setuptools.build_meta"
-
--- a/tests/test_batch.py
+++ b/tests/test_batch.py
@@ -2,10 +2,10 @@ from unittest.mock import patch
 from uuid import uuid4
 
 import pytest
-from tests.test_doubles.authentication import FakeAuth0AuthenticationSuccess
 
-from pasqal_cloud import SDK, Batch, Job
+from pasqal_cloud import Batch, Job, SDK
 from pasqal_cloud.device import BaseConfig, EmuFreeConfig, EmulatorType, EmuTNConfig
+from tests.test_doubles.authentication import FakeAuth0AuthenticationSuccess
 
 
 class TestBatch:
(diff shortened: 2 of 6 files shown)


awennersteen (approved):
lgtm

ArnaudPel:
When looking at the "unified" diff, including Augustin's initial commit plus the commits of this MR, I have a few observations:
- [minor] the argument `fetch_result` was removed from docstrings, I'd recommend keeping it and marking it as "deprecated" instead
- the contract of `Batch.declare_complete()` is still broken after this MR, because passing `fetch_results` will raise `TypeError`. I don't know whether that's accepted or not
- IMO `test_create_batch_and_fetch_results` should be kept (should be re-instated by this MR), because we still support that feature even though it's deprecated
- Do we want to edit previous changelog entries (v 0.3.0 => 0.3.2) to indicate these versions should not be used ?

To insist about the test side, since the unified diff is just supposed to deprecate a feature, if any unit-test of a public feature needs fixing, then I believe we are on the wrong track.

tests/test_batch.py
```diff
@@ -41,10 +41,14 @@ def test_create_batch(self, emulator):
...
-    def test_create_batch_and_wait(self, request_mock):
+    @pytest.mark.parametrize("wait,fetch_results", [(True, False), (False, True)])
+    def test_create_batch_and_wait(self, request_mock, wait, fetch_results):
         job = {"runs": self.n_job_runs, "variables": self.job_variables}
         batch = self.sdk.create_batch(
-            serialized_sequence=self.pulser_sequence, jobs=[job], wait=True
+            serialized_sequence=self.pulser_sequence,
+            jobs=[job],
+            wait=wait,
+            fetch_results=fetch_results,
```
ArnaudPel, line 51: Ah sorry I commented too quickly, this covers the `fetch_result` case. So I guess we can ignore my comment about testing.

pyproject.toml
```diff
@@ -1,4 +1,12 @@
+[tool.black]
```
Yoric, line 1: These changes feel unrelated, no?

Yoric (approved):

pyproject.toml
anneclairelh: Well I need these settings so that then when I run pre-commit  (which was added in the previous MR by @Mildophin ) it doesn't complain about all my crapy formatting :) Is it fine to you to add them now ?

Mildophin: mb that's something I should have added, thanks AC

Mildophin (approved):
LGTM as well

ArnaudPel (approved):

PR #95, opened 2023-07-06
[IMPROVEMENT] Batch's jobs are returned the same order as they were created + python 3.7 dropped + linting tools updated (#607) 

Mildophin:
This MR is mostly about linting following my previous MR as it wasn't applied. Also it adds a new attribute "ordered_jobs" to the Batch class which returns the jobs by their creation time. A part of the modification will be done on APIs side to make sure jobs are returned in order.

Update: This MR also includes the changes of https://github.com/pasqal-io/pasqal-cloud/pull/88, with some updates of the linting dependencies and the files of the CI and pre commit hooks. There are few things I forgot to change in my last MR, so this version patch them.

If you think it is better that I split this MR in 2, lmk, I feel guilty about mixing 3 things at once, isn't very clean.


tests/conftest.py
```diff
@@ -84,7 +84,6 @@ def batch(pasqal_client_mock):
         "start_datetime": "2023-01-01T00:00:00.000Z",
         "end_datetime": None,
         "device_status": "available",
-        "jobs": {},
```
Mildophin, line 87: was non necessary, could bring types errors later one

pasqal_cloud/batch.py
```diff
@@ -79,16 +82,28 @@ def _load_configuration(
             conf_class = EmuTNConfig
         elif values["device_type"] == EmulatorType.EMU_FREE.value:
             conf_class = EmuFreeConfig
-
         return conf_class.from_dict(configuration)
 
     @root_validator(pre=True)
-    def _build_job_dict(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+    def _build_job_dict_and_list(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        """This root validator will modify the 'jobs' attribute (which is a list
```
Mildophin, line 89: I added this docstring as it was not easy to understand how is working pydantic and root validators, easier to read for the next ones who will need to understand this as the 'jobs' transformation is something important to remember

pasqal_cloud/batch.py
```diff
@@ -112,6 +127,7 @@ def add_job(
             job_data["variables"] = variables
         job_rsp = self._client._send_job(job_data)
         job = Job(**job_rsp, _client=self._client)
+        self.ordered_jobs.append(job)
```
Mildophin, line 130: I guess this is okay, as we add jobs, it will be created after the others, so it will be at the end of the list

pasqal_cloud/batch.py
```diff
@@ -143,7 +159,20 @@ def declare_complete(
                     self.id,
                 )
             for job_rsp in batch_rsp["jobs"]:
-                self.jobs[job_rsp["id"]] = Job(**job_rsp)
+                job = Job(**job_rsp)
```
Mildophin, line 162: As ordered jobs is a list of dicts, we need a list comprehension list to be able to get the index of the batch we want to update, and then replace it by its updated data.

awennersteen:
Just FYI, in Pulser there is this open MR, let's sync
https://github.com/pasqal-io/Pulser/pull/541

.github/workflows/run-tests-and-mypy-and-black.yml
```diff
@@ -1,10 +1,6 @@
 name: Run tests and mypy
 
-on:
-  push:
-    branches: [dev]
-  pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
+on: push
```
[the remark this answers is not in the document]
awennersteen: Whatever you want tbh, just pointing it out in case you hadn't considered doing it.
And since I see we have occasionally several changes to old code in MR, presumably because the previous MR did not have all the linting and formatting applied

pasqal_cloud/batch.py
```diff
@@ -36,10 +36,12 @@ class Batch(BaseModel):
         - end_datetime: Timestamp of when the batch process was finished.
         - device_status: Status of the device where the batch is running.
         - jobs: Dictionary of all the jobs added to the batch.
+        - ordered_jobs: List of dictionaries of all the jobs added to the batch,
+            ordered by creation time.
         - jobs_count: Number of jobs added to the batch.
         - jobs_count_per_status: Number of jobs per status.
         - configuration: Further configuration for certain emulators.
-        - group_id: This parameter is deprecated, use project_id instead.
+        - group_id (deprecated): Use project_id instead.
```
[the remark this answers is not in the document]
awennersteen: Yeah, I got that :)
But I think at least one other place it was written with capital D, not sure which one we want

pasqal_cloud/__init__.py
```diff
@@ -152,15 +161,19 @@ def get_batch(self, id: str, fetch_results: bool = False) -> Batch:
         """Retrieve a batch's data and all its jobs.
 
         Args:
+            fetch_results (Deprecated): Whether to wait for the batch to be
```
awennersteen, line 164: nit: here is the capital D one

pasqal_cloud/batch.py
awennersteen: nice!

pasqal_cloud/__init__.py
Mildophin: oh yes you are right mb

.github/workflows/run-tests-and-mypy-and-black.yml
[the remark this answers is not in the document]
Mildophin: I changed it (it took a while though, i'm becoming devops haha), however i cannot configure properly the mypy hook (https://github.com/pre-commit/mirrors-mypy) to match what is checking our current configuration.

Also I updated some dependencies for dev (flake8 especially), removed the flake8 import sort, replacing it by isort to be more consistent with the backend. Also i'm thinking about opening a ticket to switch from Mypy to Pyright, for consistency aswell. Lmk what do you think @awennersteen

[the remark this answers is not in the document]
awennersteen: I went for MyPy originally because it's what mostly used outside the cloud team at Pasqal. But feel free to make it consistent with internal code!

You can now put devops engineer on your resume ;)

setup.py
```diff
@@ -28,32 +28,29 @@
...
+        "auth0-python >= 3.23.1, <4.0.0",
         "requests>=2.25.1, <3.0.0",
         "pyjwt[crypto]>=2.5.0, <3.0.0",
         "pydantic>=1.10, <2.0",
     ],
     extras_require={
         "dev": {
-            "black==22.3.0",
-            "flake8==3.9.0",
-            "flake8-import-order==0.18.1",
```
Mildophin, line 48: removed as isort is doing the job now

pasqal_cloud/batch.py
Mildophin: Updated, please let me know if it looks better, my first try was not that much readable, I think this one is better. However there might be a potential error of IndexError if there is an additional job in the job_rsp. Idk if this case if possible though

PR #97, opened 2023-07-18
[FEAT] Workload support

Guillaumemartinfesta:
This MR enables to use workloads through the SDK by using the create_workload, get_workload and cancel_workload methods which are hitting the corresponding endpoints in Core API. Unit tests and necessary fixtures were also added for that.

### DIFF
--- a/pasqal_cloud/__init__.py
+++ b/pasqal_cloud/__init__.py
@@ -87,6 +87,7 @@ class SDK:
             auth0=auth0,
         )
         self.batches: Dict[str, Batch] = {}
+        self.workloads: Dict[str,Workload] ={}
         self.webhook = webhook
 
     def create_batch(
@@ -219,23 +220,23 @@ class SDK:
                     workload_type: str,
                     backend: str,
                     config: Dict[str,Any],
-                    wait: bool = False,):
+                    wait: bool = False,)->Workload:
         req = {
             "workload_type": workload_type,
             "backend": backend,
             "config": config,
         }
-        batch_rsp = self._client._send_workload(req)
-        batch_id = batch_rsp["id"]
+        workload_rsp = self._client._send_workload(req)
+        workload_id = workload_rsp["id"]
         if wait:
             while batch_rsp["status"] in ["PENDING", "RUNNING"]:
                 time.sleep(RESULT_POLLING_INTERVAL)
-                batch_rsp = self._client._get_batch(batch_id)
+                batch_rsp = self._client._get_workload(workload_id)
 
-        batch = Batch(**batch_rsp, _client=self._client)
+        workload = Workload(**workload_rsp, _client=self._client)
 
-        self.batches[batch.id] = batch
-        return batch
+        self.workloads[workload.id] = workload
+        return workload
 
     def get_workload(self,id:str, wait: bool = False)->Workload:
         """Retrieve a workload's data.
--- a/pasqal_cloud/__init__.py
+++ b/pasqal_cloud/__init__.py
@@ -87,7 +87,7 @@ class SDK:
             auth0=auth0,
         )
         self.batches: Dict[str, Batch] = {}
-        self.workloads: Dict[str,Workload] ={}
+        self.workloads: Dict[str, Workload] = {}
         self.webhook = webhook
 
     def create_batch(
@@ -216,11 +216,13 @@ class SDK:
         job = Job(**job_rsp, _client=self._client)
         return job
 
-    def create_workload(self,
-                    workload_type: str,
-                    backend: str,
-                    config: Dict[str,Any],
-                    wait: bool = False,)->Workload:
+    def create_workload(
+        self,
+        workload_type: str,
+        backend: str,
+        config: Dict[str, Any],
+        wait: bool = False,
+    ) -> Workload:
         req = {
             "workload_type": workload_type,
             "backend": backend,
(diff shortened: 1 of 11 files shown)


pasqal_cloud/workload.py
```diff
@@ -0,0 +1,52 @@
...
+    workload_type: Optional[str] = None
+    config: Optional[Dict[str, Any]] = None
+    created_at: str
+    updated_at: str
+    errors: Optional[List[str]] = None
+    start_timestamp: Optional[str] = None
+    end_timestamp: Optional[str] = None
+    result: Optional[Dict[str, Any]] = None
+
+    class Config:
```
Mildophin, line 44: What does this exactly ? The 2 lines below

tests/test_workload.py
```diff
@@ -0,0 +1,94 @@
...
+        client_rsp = self.sdk.cancel_workload(self.workload_id)
+        assert type(client_rsp) == Workload
+        assert client_rsp.status == "CANCELED"
+        assert request_mock.last_request.method == "PUT"
+        assert (
+            request_mock.last_request.url == f"{self.sdk._client.endpoints.core}"
+            f"/api/v1/workloads/{self.workload_id}/cancel"
+        )
+
+    def test_workload_instantiation_with_extra_field(self, workload):
```
Mildophin, line 80: good idea

pasqal_cloud/workload.py
```diff
@@ -0,0 +1,52 @@
...
+        - result: Result of the workload.
+    """
+
+    id: str
+    project_id: str
+    status: str
+    _client: Client
+    backend: Optional[str] = None
+    workload_type: Optional[str] = None
+    config: Optional[Dict[str, Any]] = None
```
Augustinio, line 36: Why define a Config class below if you're going to allow an optional dict here?

Augustinio, line 36: backend, workload_type and config are not optional

pasqal_cloud/__init__.py
```diff
@@ -214,6 +216,58 @@ def cancel_job(self, id: str) -> Job:
...
+        workload_type: str,
+        backend: str,
+        config: Dict[str, Any],
+        wait: bool = False,
+    ) -> Workload:
+        req = {
+            "workload_type": workload_type,
+            "backend": backend,
+            "config": config,
+        }
```
Augustinio, line 230: I think we could create another schema to validate the payload to create the workload.
We could ensure the backend matches one of the BackendNames, and the workload_type too.
This means `WorkloadType` and `BackendName` enums should be stored in the sdk too.

I would have a `workload` directory with a `constants.py` util for storing those.
The `CreateWorkloadSchema` could also perform some validations. For example if a backend is not available for a given workload type it could be done there.

pasqal_cloud/__init__.py
```diff
@@ -214,6 +216,58 @@ def cancel_job(self, id: str) -> Job:
...
+            wait: Whether to wait for the workload to be done
+
+        Returns:
+            Job: the workload stored in the PCS database.
+        """
+        workload_rsp = self._client._get_workload(id)
+        if wait:
+            while workload_rsp["status"] in ["PENDING", "RUNNING"]:
+                time.sleep(RESULT_POLLING_INTERVAL)
+                workload_rsp = self._client._get_job(id)
```
Augustinio, line 257: `_get_workload` not `_get_job`

pasqal_cloud/__init__.py
```diff
@@ -214,6 +216,58 @@ def cancel_job(self, id: str) -> Job:
...
+            "workload_type": workload_type,
+            "backend": backend,
+            "config": config,
+        }
+        workload_rsp = self._client._send_workload(req)
+        workload_id = workload_rsp["id"]
+        if wait:
+            while workload_rsp["status"] in ["PENDING", "RUNNING"]:
+                time.sleep(RESULT_POLLING_INTERVAL)
+                workload_rsp = self._client._get_workload(workload_id)
```
Augustinio, line 236: nit: we could refactor to avoid having this code twice.


```
def _wait_for_workload(workload_resp: dict[str, Any], workload_id: str):
    while workload_rsp["status"] in ["PENDING", "RUNNING"]:
                 workload_rsp = self._client._get_job(id)
    return workload_rsp
```

Augustinio (changes requested):

pasqal_cloud/__init__.py
Guillaumemartinfesta: Both the WorkloadType and BackendName should not be disclosed in this open source directory and I thus can't make any validation on the sdk for that. When qucint will be open source we will be able to reveal (& validate) a bit more tho.

tests/test_workload.py
Guillaumemartinfesta: This comes from the unit tests for the jobs so I won't take the credit for it haha

pasqal_cloud/workload.py
Guillaumemartinfesta: This is also done in jobs, it enables to have more flexibility in the validation so that errors are not thrown if the pydantic models from public apis and pasqal cloud don't match perfectly. We could see this as an extra tool/safety measure to ensure backward compatibility.

pasqal_cloud/workload.py
Guillaumemartinfesta: Not sure to fully understand to get your point here as the Config class is just a way to configure pydantic's behavior & is not linked to the config field of the model ? By the way as suggested above this is not Optional anymore as it didn't make sense.

Augustinio: Yeah, brainfart on my side

Augustinio (approved):

PR #99, opened 2023-07-26
[FEAT] Batch's jobs are now returned in the same order they were created

Mildophin:
This MR is following changes in the APIs, from now the jobs of a Batch are returned in order of insertion in the DB.

What does this MR:
- Introduce a new Batch Attribute called 'ordered_jobs which is a list of Jobs, it is basically filled with jobs data returned by the APIs without any additional transformation
- Make Batch 'jobs' attribute deprecated
- Update current tests
- Skipped Iroise tests are now operational

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@
 All notable changes to this project will be documented in this file.
 
 
-
 ## [0.3.5] - 2023-07-21
 
 ### Added
@@ -13,7 +12,6 @@ Added an `ordered_jobs` list as an attribute to the `Batch` object in which jobs
 ### Changed
 
 Batch attribute `jobs` is now deprecated, use `ordered_jobs` instead.
-
 
 ## [0.3.4] - 2023-07-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,11 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
-Now jobs are ordered by creation date in a list through the attribute `ordered_jobs` when manipulating a Batch object.
+Added an `ordered_jobs` list as an attribute to the `Batch` object in which jobs are ordered by creation date.
+
+### Changed
+
+Batch attribute `jobs` is now deprecated, use `ordered_jobs` instead.
 
 
 ## [0.3.4] - 2023-07-18
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 All notable changes to this project will be documented in this file.
 
+
+
+## [0.3.5] - 2023-07-21
+
+### Added
+
+Now jobs are ordered by creation date in a list through the attribute `ordered_jobs` when manipulating a Batch object.
+
+
 ## [0.3.4] - 2023-07-18
 
 ### Added
--- a/pasqal_cloud/batch.py
+++ b/pasqal_cloud/batch.py
@@ -164,8 +164,9 @@ class Batch(BaseModel):
                     self.id,
                 )
 
-            jobs = [Job(**job, _client=self._client) for job in batch_rsp["jobs"]]
-            self.ordered_jobs = jobs
+            self.ordered_jobs = [
+                Job(**job, _client=self._client) for job in batch_rsp["jobs"]
+            ]
 
         return batch_rsp
 
--- a/pasqal_cloud/batch.py
+++ b/pasqal_cloud/batch.py
@@ -1,6 +1,5 @@
 import time
 from typing import Any, Dict, List, Optional, Type, Union
-from warnings import warn
 
 from pydantic import BaseModel, Extra, root_validator, validator
 
@@ -71,13 +70,13 @@ class Batch(BaseModel):
         extra = Extra.allow
         arbitrary_types_allowed = True
 
-    def __getattribute__(self, jobs):
-        warn(
-            "'jobs' attribute is deprecated, use 'ordered_jobs' instead",
-            DeprecationWarning,
-            stacklevel=1,
-        )
-        return super().__getattribute__(jobs)
+    # def __getattribute__(self, jobs):
+    #     warn(
+    #         "'jobs' attribute is deprecated, use 'ordered_jobs' instead",
+    #         DeprecationWarning,
+    #         stacklevel=1,
+    #     )
+    #     return super().__getattribute__(jobs)
 
     @validator("configuration", pre=True)
     def _load_configuration(
--- a/pasqal_cloud/batch.py
+++ b/pasqal_cloud/batch.py
@@ -69,19 +69,10 @@ class Batch(BaseModel):
         extra = Extra.allow
         arbitrary_types_allowed = True
 
-    @property
-    def jobs(self) -> dict[str, Job]:
-        warn(
-            "'jobs' attribute is deprecated, use 'ordered_jobs' instead",
-            DeprecationWarning,
-            stacklevel=1,
-        )
-        return {**{job.id: job for job in self.ordered_jobs}}
-
     @root_validator(pre=True)
     def _build_ordered_jobs(cls, values: Dict[str, Any]) -> Dict[str, Any]:
-        """This root validator will modify the 'jobs' attribute (which is a list
-        of jobs dictionaries ordered by creation time before instantiation).
+        """This root validator will modify the 'jobs' attribute which is a list
+        of jobs dictionaries ordered by creation time before instantiation.
         It will duplicate the value of 'jobs' in a new attribute 'ordered_jobs'
         to keep the jobs ordered by creation time.
         """
@@ -95,6 +86,18 @@ class Batch(BaseModel):
 
         values["ordered_jobs"] = ordered_jobs_list
         return values
+
+    @property
+    def jobs(self) -> dict[str, Job]:
+        """Once the 'ordered_jobs' is build, we need to keep the 'jobs' attribute
+        for backward compatibility with the code written by the users of the sdk
+        """
+        warn(
+            "'jobs' attribute is deprecated, use 'ordered_jobs' instead",
+            DeprecationWarning,
+            stacklevel=1,
+        )
+        return {**{job.id: job for job in self.ordered_jobs}}
 
     @validator("configuration", pre=True)
     def _load_configuration(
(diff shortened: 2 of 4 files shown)


pasqal_cloud/batch.py
```diff
@@ -66,6 +69,33 @@ class Config:
         extra = Extra.allow
         arbitrary_types_allowed = True
 
+    @property
```
Mildophin, line 72: Property used to be able to display a warning when accessing the jobs attribute

pasqal_cloud/batch.py
```diff
@@ -146,7 +164,16 @@ def declare_complete(
                     self.id,
                 )
             for job_rsp in batch_rsp["jobs"]:
-                self.jobs[job_rsp["id"]] = Job(**job_rsp)
+                job_rsp_obj = Job(**job_rsp)
+                # iterate through the ordered_job list attribute of Batch and find the
+                # index of the job received in the response to replace with updated data
+                job_index = [
+                    job_rsp_obj.id == job.id for job in self.ordered_jobs
```
MatthieuMoreau0, line 174: The order should be consistent no ?

tests/test_batch.py
```diff
@@ -72,7 +72,11 @@ def test_create_batch_and_wait(self, request_mock, wait, fetch_results):
         )  # the batch_id used in the mock data
         assert batch.sequence_builder == self.pulser_sequence
         assert batch.complete
-        assert batch.jobs
+        assert batch.ordered_jobs
+        assert batch.ordered_jobs[0].id == self.job_id
+        assert batch.ordered_jobs[0].result == self.job_result
+        assert batch.ordered_jobs[0].full_result == self.job_full_result
+        # attribute 'jobs' to be removed
```
MatthieuMoreau0, line 79: Do you want to test the warning here ?

Is the warning message going to appear at each iteration step?That might be quite annoying for the user..

pasqal_cloud/batch.py
```diff
@@ -66,6 +69,36 @@ class Config:
...
+        """Once the 'ordered_jobs' is build, we need to keep the 'jobs' attribute
+        for backward compatibility with the code written by the users of the sdk
+        """
+        warn(
+            "'jobs' attribute is deprecated, use 'ordered_jobs' instead",
+            DeprecationWarning,
+            stacklevel=1,
+        )
+        return {**{job.id: job for job in self.ordered_jobs}}
+
```
Mildophin, line 101: Should I add a setter ?

tests/test_batch.py
```diff
@@ -126,68 +130,6 @@ def test_cancel_job_sdk(self, request_mock):
             == f"{self.sdk._client.endpoints.core}/api/v1/jobs/{self.job_id}/cancel"
         )
 
-    @pytest.mark.skip(reason="Not enabled during Iroise MVP")
```
Mildophin, line 129: I'm wondering about the skipped Iroise MVP tests, they have been here for months, we can remove them right ?

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.3.4] - 2023-07-21
+
+### Added
+
+Now jobs are ordered by creation date in a list through the attribute `ordered_jobs` when manipulating a Batch object.
```
penguinoneshaw, line 9:
```suggestion
Added an `ordered_jobs` list as an attribute to the `Batch` object in which jobs are ordered by creation date.
```

pasqal_cloud/batch.py
```diff
@@ -66,6 +69,33 @@ class Config:
...
+    def jobs(self) -> dict[str, Job]:
+        warn(
+            "'jobs' attribute is deprecated, use 'ordered_jobs' instead",
+            DeprecationWarning,
+            stacklevel=1,
+        )
+        return {**{job.id: job for job in self.ordered_jobs}}
+
+    @root_validator(pre=True)
+    def _build_job_dict_and_list(cls, values: Dict[str, Any]) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
Mildophin: Yes It was renamed, I didn't see your message until now

tests/test_batch.py
Mildophin: Yes I also forgot to add it by it was planned

Mildophin: For the user, I have put a stacklevel of 1 which shouldn't spam them that much

pasqal_cloud/batch.py
Mildophin: Should be yes, if it send back all the jobs of the Batch I guess can just override the ordered jobs values by the response ?

pasqal_cloud/batch.py
penguinoneshaw: It depends - is `add_job(...)` the only supported way of adding jobs or have we, in the past, accepted adding by adding to the dict?

pasqal_cloud/batch.py
```diff
@@ -66,6 +69,36 @@ class Config:
...
+    def jobs(self) -> Dict[str, Job]:
+        """Once the 'ordered_jobs' is build, we need to keep the 'jobs' attribute
+        for backward compatibility with the code written by the users of the sdk
+        """
+        warn(
+            "'jobs' attribute is deprecated, use 'ordered_jobs' instead",
+            DeprecationWarning,
+            stacklevel=1,
+        )
+        return {**{job.id: job for job in self.ordered_jobs}}
```
penguinoneshaw, line 100: This doesn't need the kwargs spread if we're not adding the `_client` arg

penguinoneshaw (commented):
I've not checked thoroughly, these are just surface-level thoughts.

I would quite like to see the `root_validator` gone if we can manage it

pasqal_cloud/batch.py
MatthieuMoreau0: `add_job` should be the only entrypoint

pasqal_cloud/batch.py
penguinoneshaw: Can we do this without having to build the whole list first? Something like 

```python
job_index = next(n for n, job in enumerate(self.ordered_jobs) if job.id == job_rsp_obj.id)
```

CHANGELOG.md
Mildophin: I added a comment about jobs being deprecated from now

Mildophin:
@penguinoneshaw if you have any idea about how to get rid of the validator lmk, I'm also willing to make it disappear

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.3.4] - 2023-07-21
+
+### Added
+
+Added an `ordered_jobs` list as an attribute to the `Batch` object in which jobs are ordered by creation date.
```
awennersteen, line 9: Aren't Python dicts ordered in newer pythons (after 3.6 or something like that)?
So we could just keep the `jobs`? or does the db now return a list?

Mildophin: The db return a list of Jobs

tests/test_batch.py
awennersteen: shouldn't it be the opposite? Now after Iroise MVP the functionality tested by these tests should be enabled?

They might also just be so old now that it's not needed of course

pasqal_cloud/batch.py
Mildophin: The override is a bad idea ?

pasqal_cloud/batch.py
Mildophin: okay so no setter needed

tests/test_batch.py
Mildophin: Actually I was thinking about it 5 mins ago because there is a problem, some methods aren't tested at all like declare_complete or add_jobs, so I see 1 option: update the tests and make them work back cause they aren't working if we remove the skip decorator

pasqal_cloud/batch.py
MatthieuMoreau0: I don't see why it wouldn't and we could get rid of building the full list as asked by James

Mildophin:
Regarding the changes made by my MR, omw to update the skipped iroise MVP tests, it is compulsory @awennersteen

tests/test_batch.py
MatthieuMoreau0: This feature has not been properly reimplemented. Open batches are not properly supported by the dgx-connector nor the new version of pasqman. While this is not the case I don't think you should bother re-enabling these tests; you may however update the skip message as it is out of date.

AFAIK the first applications we wanted to run on QPU & emulators didn't require open batches which is why it was de-prioritized. The new workload API may support it in the near future.

Mildophin: The thing is why do we still have these features if they aren't tested ? shouldn't they be non-usable in that case ?

Mildophin: By updating the add_jobs methods I discovered that there is a problem in my code for example, so should I update the tests or should we mark these methods as non-usable ?

pasqal_cloud/batch.py
Mildophin: Actually the setter would be an option to get rid of the root validator @penguinoneshaw  no ?

Mildophin: From all the codes I tried, pydantic V1 will prevent use from using setters, so wre will keep the root validator I think

Mildophin:
@MatthieuMoreau0 @penguinoneshaw lmk if there anything to add following my changes

pasqal_cloud/batch.py
```diff
@@ -145,8 +163,13 @@ def declare_complete(
                 batch_rsp = self._client._get_batch(
                     self.id,
                 )
-            for job_rsp in batch_rsp["jobs"]:
-                self.jobs[job_rsp["id"]] = Job(**job_rsp)
+
+            jobs_list = []
```
MatthieuMoreau0, line 167: nit: Use list comprehension

tests/test_batch.py
MatthieuMoreau0: To be fair the unit tests could be re-enabled as the API response is mocked. So if it helps us catching bugs, you can give it a try

In practice the API will reject the request as batches are automatically closed for now but the message should be clear for the user

Mildophin:
All good on my side, lmk if it looks good to you.

MatthieuMoreau0 (approved):

PR #101, opened 2023-09-25
[DOC] Update of Readme 

Mildophin:
This MR fixes the readme by removing useless imports and now indicate to use .ordered-jobs which is a list of the jobs instead of .jobs which is a dict of jobs and that is now deprecated.


awennersteen (approved):
lgtm

PR #105, opened 2023-09-28
[FEAT] Add result link and download results

Augustinio:
Cf changelog

Also refactored fixture data to be version specific.


pasqal_cloud/client.py
```diff
@@ -170,7 +170,7 @@ def _send_workload(self, workload_data: Dict[str, Any]) -> Dict[str, Any]:
 
     def _get_workload(self, workload_id: str) -> Dict[str, Any]:
         workload: Dict[str, Any] = self._request(
-            "GET", f"{self.endpoints.core}/api/v1/workloads/{workload_id}"
+            "GET", f"{self.endpoints.core}/api/v2/workloads/{workload_id}"
```
Mildophin, line 173: As we are now targeting a new endpoint version wouldn't it be a breaking change ? So this wouldn't be the 0.3.6 but the 0.4 ?

pasqal_cloud/workload.py
```diff
@@ -52,3 +54,19 @@ def cancel(self) -> Dict[str, Any]:
...
+        cls, result: Optional[Dict[str, Any]], values: Dict[str, Any]
+    ) -> Optional[Dict[str, Any]]:
+        result_link: Optional[str] = values["result_link"]
+        if result or not result_link:
+            return result
+        try:
+            res = requests.get(result_link)
+            data = res.json()
+        except Exception:
+            raise ValueError("Invalid result link.")
```
Mildophin, line 69: nit comment, do we want to use a point at the end of the messages or not ? This ValueError uses one and the ValueError below doesn't.

tests/test_workload.py
```diff
@@ -2,12 +2,21 @@
...
+from pydantic import ValidationError
 
 from pasqal_cloud import SDK, Workload
 from tests.test_doubles.authentication import FakeAuth0AuthenticationSuccess
 
 
 class TestWorkload:
+    @pytest.fixture
+    def workload_with_link_id(self):
+        return "00000000-0000-0000-0000-000000000002"
```
Mildophin, line 14: Using 'str(UUID(int=0x2))' might be better as we prefer to use this instead of plain string to avoid missing a 0. :D

tests/test_workload.py
```diff
@@ -2,12 +2,21 @@
...
+
+    @pytest.fixture
+    def workload_with_invalid_link_id(self):
+        return "00000000-0000-0000-0000-000000000003"
```
Mildophin, line 18: same comment for UUID as above

pasqal_cloud/client.py
Augustinio: A breaking change is something that requires the user or consuming component to change their use of the sdk. Here the updated endpoint is in a private function and requires no action from consuming user/component. Therefore it is not a breaking change.

Mildophin:
I have re-read it and I didn't anything else to comment on this MR. 2-3 minors things to change but I approve.

Mildophin (approved):

tests/test_workload.py
```diff
@@ -1,13 +1,22 @@
...
+from pydantic import ValidationError
 
 from pasqal_cloud import SDK, Workload
 from tests.test_doubles.authentication import FakeAuth0AuthenticationSuccess
 
 
 class TestWorkload:
+    @pytest.fixture
+    def workload_with_link_id(self):
+        return UUID(int=0x2)
```
Mildophin, line 14: Might be better to convert it to string in case of it is used for other tests later on that requires it as a string

pasqal_cloud/workload.py
MatthieuMoreau0, line 69: The link could be valid but there could be a networking issue. I suggest raising a `Error while downloading workload results` error instead.

Moreover, can we create a custom exception for this like in #106 ?

tests/test_workload.py
```diff
@@ -1,13 +1,22 @@
...
 import pytest
+from pydantic import ValidationError
 
 from pasqal_cloud import SDK, Workload
 from tests.test_doubles.authentication import FakeAuth0AuthenticationSuccess
 
 
 class TestWorkload:
+    @pytest.fixture
+    def workload_with_link_id(self):
```
MatthieuMoreau0, line 13: nit: type hinting for return value missing

tests/test_workload.py
```diff
@@ -1,13 +1,22 @@
...
+        return UUID(int=0x2)
+
+    @pytest.fixture
+    def workload_with_invalid_link_id(self):
```
MatthieuMoreau0, line 17: same comment

MatthieuMoreau0 (approved):
LGTM, minor nitpicks

tests/conftest.py
```diff
@@ -34,10 +40,16 @@ def mock_core_response(request):
         return result
 
 
+def mock_result_link_response():
+    return {"some": "result"}
```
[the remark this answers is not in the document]
Augustinio: This fixture is to mock the response from the s3 fetch, not to one of our clients, I'll add details

tests/test_workload.py
```diff
@@ -88,7 +116,7 @@ def test_get_workload_error(self, mock_request_exception, workload):
         assert (
             mock_request_exception.last_request.url
             == f"{self.sdk._client.endpoints.core}"
-            f"/api/v1/workloads/{self.workload_id}"
```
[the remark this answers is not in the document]
Augustinio: I don't think they should live in the sdk. We can guarantee compatibility with previous versions of the sdk in the e2es, and keep route and integration tests in public-apis but the sdk tests should test for the current version of the sdk which no longer uses v1

pasqal_cloud/errors.py
```diff
@@ -103,6 +103,25 @@ def __init__(self) -> None:
...
+
+
+class InvalidWorkloadResultsFormatError(WorkloadException):
+    """
+    Exception class raised when download results succeeded but format
+    is not as expected.
+    """
+
+    def __init__(self, result_type: type) -> None:
+        super().__init__("Workload results should be dict but received {result_type}")
```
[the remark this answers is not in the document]
Augustinio: Yes

pasqal_cloud/workload.py
```diff
@@ -57,3 +63,19 @@ def cancel(self) -> Dict[str, Any]:
             raise WorkloadCancellingError from e
         self.status = workload_rsp.get("status", "CANCELED")
         return workload_rsp
+
+    @validator("result")
+    def result_link_to_result(
+        cls, result: Optional[Dict[str, Any]], values: Dict[str, Any]
+    ) -> Optional[Dict[str, Any]]:
+        result_link: Optional[str] = values["result_link"]
```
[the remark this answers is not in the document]
Augustinio: Good catch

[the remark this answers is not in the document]
Augustinio: Actually I don't think there'll be a KeyError. The Key is defined no matter what in the pydantic schema. Tests don't raise an error.

Mildophin (approved):

MatthieuMoreau0 (approved):
Ok with the changes, I'll let @oliver-gordon  confirm his comments have been properly addressed.

PR #106, opened 2023-10-02
[FIX] Client - Handle client errors

Augustinio:
Cf changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,19 @@ All notable changes to this project will be documented in this file.
 
 ## [0.3.6] - 2023-10-02
 
+### Added
+
+- Added exception classes for all possible failures (mostly related to client errors).
+- Added try-catch to corresponding methods to raise proper error
+
 ### Changed
 
-Use `raise_for_status` on response from client before returning `data` to get accurate exception.
+- Use `raise_for_status` on response from client before returning `data` to get accurate exception.
+- Bumped major as new exceptions are raised
+
+### Removed
+
+- Removed obsolete `HTTPError` and `LoginError` classes.
 
 ## [0.3.5] - 2023-07-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Use `raise_for_status` on response from client before returning `data` to get accurate exception.
-- Bumped major as new exceptions are raised
+- Bumped minor as new exceptions are raised
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.3.6] - 2023-10-02
+
+### Changed
+
+Use `raise_for_status` on response from client before returning `data` to get accurate exception.
 
 ## [0.3.5] - 2023-07-21
 
--- a/pasqal_cloud/__init__.py
+++ b/pasqal_cloud/__init__.py
@@ -105,13 +105,11 @@ class SDK:
     def _get_batch(
         self,
         id: str,
-    ):
+    ) -> Dict[str, Any]:
         try:
             return self._client._get_batch(id)
         except HTTPError as e:
-            raise BatchFetchingError(
-                f"Couldn't fetch batch data due to a client error: {e}"
-            )
+            raise BatchFetchingError(e)
 
     def create_batch(
         self,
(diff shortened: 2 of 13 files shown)


pasqal_cloud/client.py
```diff
@@ -108,9 +107,8 @@ def _request(
             headers={"content-type": "application/json"},
             auth=self.authenticator,
         )
+        rsp.raise_for_status()
         data: JSendPayload = rsp.json()
-        if rsp.status_code >= 400:
-            raise HTTPError(data)
```
MatthieuMoreau0, line 113: From what I can tell, it looks like this error is now unused in our codebase. There is only one extra reference where we catch it but it is not raised so it looks we should get rid of it

pasqal_cloud/client.py
```diff
@@ -108,9 +107,8 @@ def _request(
             headers={"content-type": "application/json"},
             auth=self.authenticator,
         )
+        rsp.raise_for_status()
```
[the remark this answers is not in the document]
Augustinio: Can do!

CHANGELOG.md
```diff
@@ -4,9 +4,19 @@ All notable changes to this project will be documented in this file.
...
+### Added
+
+- Added exception classes for all possible failures (mostly related to client errors).
+- Added try-catch to corresponding methods to raise proper error
+
 ### Changed
 
-Use `raise_for_status` on response from client before returning `data` to get accurate exception.
+- Use `raise_for_status` on response from client before returning `data` to get accurate exception.
+- Bumped major as new exceptions are raised
```
MatthieuMoreau0, line 15: Bumped minor* 

 Major would be v1.0.0

pasqal_cloud/__init__.py
```diff
@@ -143,12 +170,18 @@ def create_batch(
         if configuration:
             req.update({"configuration": configuration.to_dict()})  # type: ignore
 
-        batch_rsp = self._client._send_batch(req)
+        try:
+            batch_rsp = self._client._send_batch(req)
+        except HTTPError as e:
+            raise BatchCreationError(
+                f"Couldn't create batch due to a client error: {e}"
+            )
```
MatthieuMoreau0: Why do we sometimes pass the exception directly and sometimes pass a formatted string? Can we make it consistent?

tests/test_batch.py
```diff
@@ -82,26 +99,46 @@ def test_create_batch_and_wait(self, request_mock, wait, fetch_results):
...
         assert len(batch.ordered_jobs) == 2
 
-    def test_batch_add_job_and_wait_for_results(self, request_mock):
+    def test_batch_add_job_failure(self, request_mock, request_mock_exception):
+        request_mock.start()
+        batch = self.sdk.create_batch(
+            serialized_sequence=self.pulser_sequence, jobs=[self.simple_job_args]
+        )
+        request_mock.stop()
+        request_mock_exception.start()
```
MatthieuMoreau0, line 128: Looks like the purpose of this is to have a Batch object on which we can call the `add_job` method. Maybe we could simply create an instance manually (there might even be a fixture that does just that already)

tests/test_batch.py
```diff
@@ -116,78 +153,145 @@ def test_batch_add_job_and_wait_for_results(self, request_mock):
...
         rsp = batch.declare_complete(wait=False)
         assert rsp["complete"]
 
-    def test_batch_declare_complete_and_wait_for_results(self, request_mock):
+    def test_batch_declare_complete_failure(self, request_mock, request_mock_exception):
+        request_mock.start()
+        batch = self.sdk.create_batch(
+            serialized_sequence=self.pulser_sequence, jobs=[self.simple_job_args]
+        )
+        request_mock.stop()
```
MatthieuMoreau0, line 177: same question

MatthieuMoreau0 (commented):
Overall good change, thanks for taking it the time of applying it everywhere. 

On a general note, I'm wondering if casting to a string the underlying error is the best way to go about it. I'm not 100% sure how it would look like effectively in user scripts and I'm a bit worried we may lose a lot of information in that process. Do you have an example of what it looks like ?

I'm wondering if doing instead `raise CustomException from e` would not be a better way to go about this; to make sure the full stack trace is still accessible if needed. For example for validation errors returned by the API; is it clear in the stacktrace  which attribute failed validationwith the current implementation?

pasqal_cloud/__init__.py
Augustinio: That's on an outdated version of the code, we do constitently pass the exception.

Mildophin (approved):
Tests are better defined this way, the error classification is also better to know what is going on when there is an error. Getting rid of calling ._client. every time too is appreciable. Approved.

MatthieuMoreau0:
LGTM

MatthieuMoreau0 (approved):

PR #107, opened 2023-10-10
[IMPROVEMENT] Add context to exceptions

Augustinio:
Cf changelog.


pasqal_cloud/errors.py
```diff
@@ -1,4 +1,18 @@
...
+
+class ExceptionWithResponseContext(BaseException):
+    def __init__(self, msg: str, e: HTTPError = None) -> None:
+        if not e:
+            return super().__init__(msg)
+        data = "without context."
+        resp: Response = e.response
+        if resp.content:
+            data = resp.json()
+        super().__init__(f"{msg}: {data}")
```
MatthieuMoreau0, line 12: Before the error rework, we had custom logic to parse the response body (something like look at `data` attribute if any, otherwise print the `message`). How does this compare to it?

MatthieuMoreau0 (commented):
Can we add a test with a real world error response e.g. for batch creation? This would give me a bit more confidence and could serve as documentation for future references on how we want to format our error messages

pasqal_cloud/errors.py
Augustinio: I''ll refactor to mirror what was done

MatthieuMoreau0 (approved):

PR #108, opened 2023-11-06
[FEAT] Add jobs to an open batch

MatthieuMoreau0:
### Description

Let users add several jobs at once to a batch. I got rid of the previous function which let users add one job to a batch; but this was very inefficient as it required one http call per job. This is technically a breaking change but it couldn't be used by our users since the dedicated endpoint was always returning an error because batches were automatically marked as complete. Hence I think this can be freely released without worrying about breaking existing pipelines for our users.

Changes:

- Let users create an open batch
- Let users add jobs to an open batch
- Add some documentation on how to create an open batch, add jobs to it and complete it. 
- Update changelog
- Update version

### Checklist

#### Documentation

#### Tests

- [x] Tests were run locally.


pasqal_cloud/batch.py
```diff
@@ -120,40 +120,41 @@ def _load_configuration(
...
+        ]
         if wait:
-            while job.status in ["PENDING", "RUNNING"]:
+            while any(
+                [job.status in {"PENDING", "RUNNING"} for job in self.ordered_jobs]
+            ):
                 time.sleep(RESULT_POLLING_INTERVAL)
                 try:
-                    job_rsp = self._client._get_job(job.id)
+                    batch_rsp = self._client._get_batch(self.id)
```
MatthieuMoreau0, line 152: This should update the batch as well

pasqal_cloud/__init__.py
```diff
@@ -151,11 +155,16 @@ def create_batch(
                 DeprecationWarning,
                 stacklevel=2,
             )
+            wait = wait or fetch_results
+
+        if wait and not complete:
+            raise ValueError("Cannot wait for an open batch.")
```
Augustinio, line 161: In `add_jobs` the behaviour of wait is to wait until all jobs are in termination status. Here it is to wait until the batch is set to done.

Wouldn't it be more consistent to change the behaviour of the wait on open (and closed) batches to also depend on the status of all jobs?

Augustinio (commented):
LGTM, just one question

pasqal_cloud/batch.py
```diff
@@ -120,45 +120,42 @@ def _load_configuration(
...
                 except HTTPError as e:
                     raise JobFetchingError(e) from e
-                job = Job(**job_rsp)
-        return job
+                batch = Batch(**batch_rsp, _client=self._client)
```
MatthieuMoreau0, line 152: @MohamedAminMallek @Augustinio I had to change the interface; now the function returns a new instance instead of modifying in place the batch object. 

Ideally I would prefer to update the object in place but the previous implementation which set `self=Batch(...)` doesn't actually update the current object. We could update all attributes one by one based on the http response but I haven't found a nice pattern just yet. I'm open to ideas

MohamedAminMallek: I find it a little strange that a batch instance method doesn't update the state of the instance and return a new one. Perhaps we could have a dedicated private method that takes a dictionary (HTTP response) and updates the state of the batch.


Out of scope:
Why do we have a cancel_batch method in the SDK client and another one in the batch class?

MatthieuMoreau0: Yeah I dislike it as well. I'll see what I can do.

>Why do we have a cancel_batch method in the SDK client and another one in the batch class?
It's mostly for convenience for the user to make it possible to cancel a batch whether one has a Batch instance or simply the batch_id at hand

pasqal_cloud/__init__.py
MatthieuMoreau0: I updated it following your suggestion.

pasqal_cloud/batch.py
MatthieuMoreau0: I introduced a `refresh` method and used to it to update the object in place in the different methods, let me know if that looks good to you!

PR #115, opened 2024-04-09
[IMPROVEMENT] Add exponential backoff retries to HTTP client

oliver-gordon:
### Description

When a HTTP request fails, we will now retry it 5 times (arbitrary value, non-configurable currently) where the time between each request grows until we stop.

I've also tried to include for missing typehints and docstrings where I was updating code.

### Checklist

#### Documentation

#### Tests

- [ ] Tests were run locally.

### DIFF
--- a/pasqal_cloud/client.py
+++ b/pasqal_cloud/client.py
@@ -111,7 +111,7 @@ class Client:
     ) -> JSendPayload:
         successful_request: bool = False
         iteration: int = 0
-        while not successful_request and iteration <= HTTP_RETRIES:
+        while iteration <= HTTP_RETRIES and not successful_request:
             # time = (interval seconds * exponent rule) ^ retries
             delay = (1 * 2) ** iteration
             rsp = requests.request(
@@ -125,6 +125,7 @@ class Client:
             )
             try:
                 rsp.raise_for_status()
+                successful_request = True
             except Exception as e:
                 if (
                     rsp.status_code not in {408, 425, 429, 500, 502, 503, 504}
@@ -132,10 +133,12 @@ class Client:
                 ):
                     raise e
 
+            if successful_request:
+                data: JSendPayload = rsp.json()
+                return data
+
             time.sleep(delay)
             iteration += 1
-        data: JSendPayload = rsp.json()
-        return data
 
     def _send_batch(self, batch_data: Dict[str, Any]) -> Dict[str, Any]:
         batch_data.update({"project_id": self.project_id})
--- a/pasqal_cloud/client.py
+++ b/pasqal_cloud/client.py
@@ -18,7 +18,6 @@ from getpass import getpass
 from typing import Any, Dict, List, Mapping, Optional, Sequence, Union
 import time
 import requests
-from requests import HTTPError
 from requests.auth import AuthBase
 
 from pasqal_cloud.authentication import (
@@ -126,11 +125,13 @@ class Client:
                     params=params,
                 )
                 data: JSendPayload = rsp.json()
+                successful_request = True
             except Exception as e:
                 if iteration == HTTP_RETRIES:
-                    raise HTTPError
+                    raise e
                 time.sleep(delay)
             iteration += 1
+
         return data
 
     def _send_batch(self, batch_data: Dict[str, Any]) -> Dict[str, Any]:
--- a/pasqal_cloud/client.py
+++ b/pasqal_cloud/client.py
@@ -115,7 +115,6 @@ class Client:
         while not successful_request and iteration <= HTTP_RETRIES:
             # time = (interval seconds * exponent rule) ^ retries
             delay = (1 * 2) ** iteration
-            # resp: requests.Request = None
             try:
                 rsp = requests.request(
                     method,
@@ -231,9 +230,6 @@ class Client:
         workload: Dict[str, Any] = self._request(
             "PUT", f"{self.endpoints.core}/api/v1/workloads/{workload_id}/cancel"
         )
-        print("*" * 100)
-        print(workload)
-        print("*" * 100)
         return workload["data"]
 
     def get_device_specs_dict(self) -> Dict[str, str]:
--- a/pasqal_cloud/client.py
+++ b/pasqal_cloud/client.py
@@ -134,7 +134,8 @@ class Client:
 
             time.sleep(delay)
             iteration += 1
-        return rsp.json()
+        data: JSendPayload = rsp.json()
+        return data
 
     def _send_batch(self, batch_data: Dict[str, Any]) -> Dict[str, Any]:
         batch_data.update({"project_id": self.project_id})
--- a/pasqal_cloud/client.py
+++ b/pasqal_cloud/client.py
@@ -114,25 +114,27 @@ class Client:
         while not successful_request and iteration <= HTTP_RETRIES:
             # time = (interval seconds * exponent rule) ^ retries
             delay = (1 * 2) ** iteration
+            rsp = requests.request(
+                method,
+                url,
+                json=payload,
+                timeout=TIMEOUT,
+                headers={"content-type": "application/json"},
+                auth=self.authenticator,
+                params=params,
+            )
             try:
-                rsp = requests.request(
-                    method,
-                    url,
-                    json=payload,
-                    timeout=TIMEOUT,
-                    headers={"content-type": "application/json"},
-                    auth=self.authenticator,
-                    params=params,
-                )
-                data: JSendPayload = rsp.json()
-                successful_request = True
+                rsp.raise_for_status()
             except Exception as e:
-                if iteration == HTTP_RETRIES:
+                if (
+                    rsp.status_code not in {408, 425, 429, 500, 502, 503, 504}
+                    or iteration == HTTP_RETRIES
+                ):
                     raise e
-                time.sleep(delay)
-            iteration += 1
 
-        return data
+            time.sleep(delay)
+            iteration += 1
+        return rsp.json()
 
     def _send_batch(self, batch_data: Dict[str, Any]) -> Dict[str, Any]:
         batch_data.update({"project_id": self.project_id})
--- a/pasqal_cloud/client.py
+++ b/pasqal_cloud/client.py
@@ -16,8 +16,9 @@ from __future__ import annotations
 from datetime import datetime
 from getpass import getpass
 from typing import Any, Dict, List, Mapping, Optional, Sequence, Union
-
+import time
 import requests
+from requests import HTTPError
 from requests.auth import AuthBase
 
 from pasqal_cloud.authentication import (
@@ -29,6 +30,7 @@ from pasqal_cloud.endpoints import Auth0Conf, Endpoints
 from pasqal_cloud.utils.jsend import JSendPayload
 
 TIMEOUT = 30  # client http requests timeout after 30s
+HTTP_RETRIES = 5  # HTTP client will raise an exception after too many consecutive fails
 
 
 class EmptyFilter:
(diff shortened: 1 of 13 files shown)


tests/test_batch.py
```diff
@@ -183,48 +236,52 @@ def custom_get_batch_mock(request, _):
...
-        assert len(batch.ordered_jobs) == 2
-        assert (
-            mock_request.call_count == 5
-        )  # 1 call to add jobs and 4 get batch calls until jobs are DONE
+        assert mock_request.call_count == 12
+
         assert (
             mock_request.request_history[0].url
-            == f"{self.sdk._client.endpoints.core}/api/v1/batches/{batch.id}/jobs"
+            == f"{self.sdk._client.endpoints.core}/api/v1/batches/00000000-0000-0000-0000-000000000002/jobs"
```
oliver-gordon, line 241: There was some unsavory mocked behavior I could not pinpoint, but compared IDs were not equal. Really struggled to pinpoint why

tests/test_batch.py
```diff
@@ -183,48 +236,52 @@ def custom_get_batch_mock(request, _):
...
         )
+
         assert mock_request.request_history[0].method == "POST"
-        assert all(
-            [
-                (req.method, req.url)
-                == (
-                    "GET",
```
oliver-gordon, line 204: This became an invalid assertion so I've removed it in favor of a lesser check

pasqal_cloud/errors.py
```diff
@@ -8,7 +8,12 @@ class ExceptionWithResponseContext(BaseException):
     def __init__(self, msg: str, e: Optional[HTTPError] = None) -> None:
         if not e:
             return super().__init__(msg)
+
+        if e.response is None:
```
oliver-gordon, line 12: There is a mock library used in the SDK tests that when returning an exception instead of a payload, does not return response objects, so this change was introduced, which shouldn't provide issues with standard use

pasqal_cloud/client.py
```diff
@@ -108,18 +110,27 @@ def _request(
...
-            json=payload,
-            timeout=TIMEOUT,
-            headers={"content-type": "application/json"},
-            auth=self.authenticator,
-            params=params,
-        )
-        rsp.raise_for_status()
-        data: JSendPayload = rsp.json()
-
+        successful_request: bool = False
```
oliver-gordon, line 113: There is currently no override, I don't know if we want this to be forced default behavior (you would not see it take place if no retry happens) or if we want to make it configurable with an argument. I'd argue for the override arg but can't think of a scenario you want to

tests/test_batch.py
```diff
@@ -183,48 +224,52 @@ def custom_get_batch_mock(request, _):
 
         # Reset history so that we can count calls later
         mock_request.reset_mock()
+        mock_request.register_uri(
+            "POST",
+            f"/core-fast/api/v1/batches/{batch.id}/jobs",
+            json=load_mock_batch_json_response,
```
MatthieuMoreau0, line 230: This should use the POST `/core-fast/api/v1/batches/00000000-0000-0000-0000-000000000002/jobs` fixture instead of the POST /batches fixture. 

I believe this is the reason why the assertion `mock_request.request_history[0].url == f"{self.sdk._client.endpoints.core}/api/v1/batches/{batch.id}/jobs"` stopped working

pasqal_cloud/client.py
```diff
@@ -218,8 +229,8 @@ def _get_workload(self, workload_id: str) -> Dict[str, Any]:
     def _cancel_workload(self, workload_id: str) -> Dict[str, Any]:
         workload: Dict[str, Any] = self._request(
```
awennersteen, line 230:
```suggestion
        workload: JSendPayload= self._request(
```

tests/test_batch.py
```diff
@@ -267,7 +318,16 @@ def test_cancel_batch_sdk(self, mock_request):
             f"/api/v1/batches/{self.batch_id}/cancel"
         )
 
-    def test_cancel_batch_sdk_error(self, mock_request_exception):
+    def test_cancel_batch_sdk_error(
+        self, mock_request_exception: Generator[Any, Any, None]
+    ):
+        """
+        Assert that a BatchCancellingError is raised appropriatrely for
+        failling requests when calling sdk.cancel_batch(...)
```
MatthieuMoreau0, line 326:
```suggestion
        Assert that a BatchCancellingError is raised appropriately for
        failing requests when calling sdk.cancel_batch(...)
```

tests/test_batch.py
```diff
@@ -74,30 +93,32 @@ def test_create_batch(self, emulator, mock_request):
...
+        and that the correct methods and URLs are used.
+        """
+        with pytest.raises(BatchCreationError):
             _ = self.sdk.create_batch(
-                serialized_sequence=self.pulser_sequence,
-                jobs=[self.simple_job_args],
+                serialized_sequence=self.pulser_sequence, jobs=[self.simple_job_args]
             )
 
+        assert mock_request_exception.last_request.method == "POST"
```
MatthieuMoreau0, line 108: In this test we could assert that we retried the API up to the max retry value

MatthieuMoreau0 (commented):
Thanks for adding a lot of missing typehints and docstrings. I pointed out a few typos. I have some other open questions & suggestions I'd like to discuss before this is merged

pasqal_cloud/client.py
oliver-gordon: I think this was a change I forgot to revert  which is why it's alone

awennersteen: cool, makes sense.
Figured that once I saw it I'd save you another few moments living in Python land, since it was what broke your mypy ;)

tests/test_batch.py
oliver-gordon: I don't understand this comment. Do you mean this fixture?  `json=load_mock_batch_json_response`

tests/test_batch.py
```diff
@@ -183,48 +224,52 @@ def custom_get_batch_mock(request, _):
...
         batch.add_jobs(
             [{"runs": self.n_job_runs}, {"runs": self.n_job_runs}],
             wait=True,
         )
 
-        assert len(batch.ordered_jobs) == 2
-        assert (
-            mock_request.call_count == 5
-        )  # 1 call to add jobs and 4 get batch calls until jobs are DONE
+        assert mock_request.call_count == 12
```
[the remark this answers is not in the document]
oliver-gordon: Well I've included the mock_request for POST, which was missing previously, so I believe it's accounting for those

pasqal_cloud/client.py
```diff
@@ -108,18 +110,27 @@ def _request(
...
-            headers={"content-type": "application/json"},
-            auth=self.authenticator,
-            params=params,
-        )
-        rsp.raise_for_status()
-        data: JSendPayload = rsp.json()
-
+        successful_request: bool = False
+        iteration: int = 0
+        while not successful_request and iteration <= HTTP_RETRIES:
```
oliver-gordon, line 115: Looks like I forgot to update `successful_request` when the request returns a payload

pasqal_cloud/client.py
```diff
@@ -108,17 +109,28 @@ def _request(
...
+                    headers={"content-type": "application/json"},
+                    auth=self.authenticator,
+                    params=params,
+                )
+                data: JSendPayload = rsp.json()
+                successful_request = True
+            except Exception as e:
+                if iteration == HTTP_RETRIES:
+                    raise e
+                time.sleep(delay)
```
ArnaudPel, line 132: Do we want to differentiate between retryable vs non-retryable error codes ? https://stackoverflow.com/questions/47680711

For example, re-sending a "bad request" is unlikely to make it work, more likely something needs to be fixed on the client side.

oliver-gordon: Yeah I agree with this, I suppose it's a rather small pool of status codes that we want to retry for. It's mostly going to be 502s of 408s I imagine

MatthieuMoreau0: 500s as well given the status code currently returned by our APIs

pasqal_cloud/client.py
```diff
@@ -108,17 +109,28 @@ def _request(
         payload: Optional[Union[Mapping, Sequence[Mapping]]] = None,
```
MohamedAminMallek, line 109: Why not use a dedicated library for retry logic? 
Something like [tenacity](https://tenacity.readthedocs.io/en/latest/)

pasqal_cloud/client.py
MatthieuMoreau0: This is still missing AFAICT

tests/test_batch.py
MatthieuMoreau0: THis looks OK to me now. idk if it was changed or what I had in mind, resolving.

tests/test_batch.py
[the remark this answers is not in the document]
MatthieuMoreau0: That endpoint is called only once so at most this should have increased the call count by one. I think instead the request is retried no matter the status code. This comes from the `succesful_request` flag not being updated as you mentioned in another comment

MatthieuMoreau0 (changes requested):
It looks like the retry is performed even in success case

pasqal_cloud/client.py
oliver-gordon: Why add another 3rd party dependency when the logic is rather simple

pasqal_cloud/client.py
oliver-gordon: I know? I've not asked for a re review  yet

MatthieuMoreau0: Ok sure. Just made a pass to checkout your answers to my comments as I hadn't done so already :)

tests/test_workload.py
```diff
@@ -42,7 +43,13 @@ def init_sdk(self):
         self.config = {"test1": "test1", "test2": 2}
         self.workload_result = {"1001": 12, "0110": 35, "1111": 1}
 
-    def test_create_workload(self, mock_request):
+    @pytest.fixture(autouse=True)
+    def mock_sleep(self):
+        """Fixture to mock sleep to make tests faster."""
+        with patch("time.sleep"):
```
Mildophin, line 49: How does it work exactly ? How it makes tests faster ? I don't understand.

oliver-gordon: If you patch all `time.sleep` commands, any behavior that is delayed due to a process sleeping will ignore the delay.

Specifically in this MR, 5 HTTP retries would take 31 seconds in total.

I did not write that docstr, it's a copied fixture

pasqal_cloud/client.py
MatthieuMoreau0: Suggestion: We could make the number of retries an optional argument of this function. This would both give us the ability to control the number of retries or disable retries altogether by setting it to 0. 

This is not required for now, just a thought

pasqal_cloud/client.py
```diff
@@ -108,19 +109,42 @@ def _request(
...
+                    rsp.status_code not in {408, 425, 429, 500, 502, 503, 504}
+                    or iteration == HTTP_RETRIES
+                ):
+                    raise e
+
+            if successful_request:
+                data: JSendPayload = rsp.json()
+                return data
+
+            time.sleep(delay)
```
MatthieuMoreau0, line 140: Should we log to the user that the request failed and we are retrying?

tests/test_errors.py
```diff
@@ -7,24 +7,38 @@
 
 
 def test_exception_for_response_with_description():
+    """
+    When we catch an exception with a populated Response object,
+    the error message should extract data from the object
+    and format the message. Then we assert it's as expected
+    with an f-string pattern.
```
MatthieuMoreau0, line 14: nit: imo "Then we assert it's as expected with an f-string pattern" is a test implementation detail and doesn't provide relevant information to the reader about what the test does or why and thus could be removed. Same comment for other tests.

pasqal_cloud/client.py
MohamedAminMallek: Okay, if you prefer implementing the feature, let's do it. 

But maybe we could draw inspiration from them and separate the retry logic from the method we want to retry? Or do you think it's not worth it since we're only doing it in one place?

MatthieuMoreau0 (commented):
I couldn't find a test for the retry mechanism, could you add one?

tests/test_workload.py
Mildophin: Alright, thx for giving me details 👍🏻

pasqal_cloud/client.py
oliver-gordon: I'm not a fan of adding logs to SDKs, they should be silent in the run time

tests/test_errors.py
oliver-gordon: You mean opposed to the nothing that was there before?

oliver-gordon:
> I couldn't find a test for the retry mechanism, could you add one?

Any test mocking an exception is testing the retry mechanism. But sure, I'll add another one

tests/test_errors.py
MatthieuMoreau0: Obviously this is infinitely better than no docstring; I never argued otherwise. I just think this specific part doesn't bring much

MatthieuMoreau0:
> > I couldn't find a test for the retry mechanism, could you add one?
> 
> Any test mocking an exception is testing the retry mechanism. But sure, I'll add another one

Either adding a new test or amending the existing ones to check that the request was retried as expected. I could get rid of the retry logic entirely and tests would still pass which is no good.

pasqal_cloud/client.py
MatthieuMoreau0: Yeah they should be silent by default I agree. My recommendation would be to add those log statement but configure the logger so that they do not show up unless the user changes the logger configuration explicitly in his script. Anyway this is outside the scope of this MR, so I'll resolve

MatthieuMoreau0 (approved):
Thanks for adding the retry tests, lgtm

PR #121, opened 2024-05-16
[FEAT] Get list of jobs

Mildophin:
### Description

This PR introduces a new method, get_jobs, for retrieving a list of jobs from the Pasqal backend using the V2 endpoint. The method utilizes 2 Pydantic class:
- JobFilters, allowing users to filter jobs without managing pagination manually
- PaginationParams, allowing users to manipulate pagination details such as the offset and the limit of jobs per page.

Additionally, similar enhancements have been applied to the rebatch method, incorporating the RebatchFilters class for filtering. This change is backward-incompatible.

Tests have been updated to match the new logic.

##### New Method: get_jobs

- Retrieves a list of jobs from the Pasqal backend.
- Filters jobs using the JobFilters and PaginationParams Pydantic classes.
- Does not return results from jobs.

##### Filtering Capabilities

Users can filter jobs based on several fields:

- id
- project_id
- user_id
- batch_id
- status
- min_runs
- max_runs
- errors
- origin
- start_date
- end_date

Note: Filtering by id__starts_with is not supported.

##### Pagination Management

Pagination is handled by the users with the class PaginationParams. It is composed of 2 fields:

- offset: An integer representing the starting index of the data to return.
- limit: An integer representing the maximum number of data entries to return.

##### Breaking Change: rebatch Method

The rebatch method now uses the RebatchFilters class for filtering, similar to get_jobs.
Users must adapt to this new filtering mechanism.

#### Remaining Tasks

Adds updated E2E tests.

#### Additional merge criteria

The E2E tests need to pass.

### Checklist

#### Documentation

#### Tests

- [x] Tests were run locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.11.0] - 2024-06-14
+
+### Added
+
+- Introduced a new filtering system using Pydantic classes for more flexible and robust filtering options
+- Added 'get_jobs' method, a feature to retrieve a list of jobs based on 'JobFilters'
+
+### Breaking change
+
+- Updated the 'rebatch' method to use 'RebatchFilters' for filtering jobs to retry,
+  replacing the previous multiple filter parameters
+
 ## [0.10.1] - 2024-06-14
 
 ### Changed
@@ -16,11 +28,6 @@ All notable changes to this project will be documented in this file.
 - Expose add_jobs and close_batch functions in the SDK interface
 - Refactor non-private methods that were prefixed with _
 - Add test coverage for new functions
-
-## [0.9.0] - 2024-06-05
-
-### Changed
-
 - Drop priority from Batch attributes
 
 ## [0.9.0] - 2024-05-15
@@ -70,8 +77,8 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - Added feature to create an "open" batch.
-  - To create an open batch, set the `complete` argument to `True` in the `create_batch` method of the SDK.
-  - To add jobs to an open batch, use the `add_jobs` method.
+    - To create an open batch, set the `complete` argument to `True` in the `create_batch` method of the SDK.
+    - To add jobs to an open batch, use the `add_jobs` method.
 - Updated documentation to add examples to create open batches.
 - The `wait` argument now waits for all the jobs to be terminated instead of waiting for the batch to be terminated.
 
@@ -182,7 +189,8 @@ hence the feature was not documented.
 
 ### Changed
 
-Fixed incorrect type hint of user IDs from int to str which was raising a validation exception when loading the data returned by the API.
+Fixed incorrect type hint of user IDs from int to str which was raising a validation exception when loading the data
+returned by the API.
 
 ## [0.2.7] - 2023-06-08
 
@@ -193,8 +201,11 @@ Now you can get the environment configuration for `endpoints` and `auth0` of the
 
 ### Changed
 
-- Batch and Job dataclasses have been replaced by pydantic models, which gives more control to unserialize the API response data.
-  The SDK is now more resilient to API changes, if a new field is added to the response of the API then the job and batch object instantiation will not raise an exception, meaning this SDK version will not become obsolete as soon as the API spec is updated.
+- Batch and Job dataclasses have been replaced by pydantic models, which gives more control to unserialize the API
+  response data.
+  The SDK is now more resilient to API changes, if a new field is added to the response of the API then the job and
+  batch object instantiation will not raise an exception, meaning this SDK version will not become obsolete as soon as
+  the API spec is updated.
 
 ## [0.2.6] - 2023-05-29
 
@@ -276,8 +287,10 @@ Note that for backwards compatibility, the `group_id` is still exposed by the AP
 - The default values for the tensor network emulator were updated to better ones.
 - the client_id and client_secret was leftover in the Client object even though they are no longer used.
 - Updated the README to also supply the group_id which is mandatory.
-- Updated the default endpoint to `/core-fast` in accordance with infra changes. All users should use `/core-fast` in all environments.
-- The PCS APIs Client was refactored to accept any custom token provider for authentication. This can be used by users as an alternative to the username/password-based token provider.
+- Updated the default endpoint to `/core-fast` in accordance with infra changes. All users should use `/core-fast` in
+  all environments.
+- The PCS APIs Client was refactored to accept any custom token provider for authentication. This can be used by users
+  as an alternative to the username/password-based token provider.
 
 ## [0.1.12] - 2023-02-27
 
@@ -299,9 +312,12 @@ Note that for backwards compatibility, the `group_id` is still exposed by the AP
 
 ### Changed
 
-- A new device type, "EMU-TN", corresponding to tensor network-based emulator, was added. The "EMU_SV" type was removed as it is not available right now.
+- A new device type, "EMU-TN", corresponding to tensor network-based emulator, was added. The "EMU_SV" type was removed
+  as it is not available right now.
 
-- A new version of the "core" microservice, using FastAPI instead of Flask as web framework, was released in the "dev" environment. If using the "dev" environment, you should upgrade the core endpoint you are using to "https://apis.dev.pasqal.cloud/core-fast"
+- A new version of the "core" microservice, using FastAPI instead of Flask as web framework, was released in the "dev"
+  environment. If using the "dev" environment, you should upgrade the core endpoint you are using
+  to "https://apis.dev.pasqal.cloud/core-fast"
 
 ## [0.1.9] - 2023-02-07
 
@@ -315,25 +331,37 @@ Note that for backwards compatibility, the `group_id` is still exposed by the AP
 ### Changed
 
 - Moved the device_types into device module
-- Refactored configuration to be split into `BaseConfig`, `EmuSVConfig` and `EmuFreeConfig`, more device-specific configs can be added
+- Refactored configuration to be split into `BaseConfig`, `EmuSVConfig` and `EmuFreeConfig`, more device-specific
+  configs can be added
 - Refactored unit tests to use the proper Config model.
 - Updated README with the new Configuration classes.
 
 ### Added
 
-- `BaseConfig`: the base configuration class. A dataclass with the same methods as the former `Configuration` model and the `extra_config` param.
-- `EmuSVConfig`: the configuration class for `DeviceType.EMU_SV`, inherits from `BaseConfig` with the parameters formerly found on `Configuration`.
-- `EmuFreeConfig`: the configuration class for `DeviceType.EMU_FREE`, inherits from `BaseConfig` with the `with_noise` boolean parameter.
+- `BaseConfig`: the base configuration class. A dataclass with the same methods as the former `Configuration` model and
+  the `extra_config` param.
+- `EmuSVConfig`: the configuration class for `DeviceType.EMU_SV`, inherits from `BaseConfig` with the parameters
+  formerly found on `Configuration`.
+- `EmuFreeConfig`: the configuration class for `DeviceType.EMU_FREE`, inherits from `BaseConfig` with the `with_noise`
+  boolean parameter.
 
 ## [0.1.7] - 2023-01-04
 
 ### Changed
 
-Reworked the `wait` logic when [creating a batch](https://github.com/pasqal-io/pasqal-cloud/blob/dev/sdk/__init__.py#L46) or [declaring it as complete](<(https://github.com/pasqal-io/pasqal-cloud/blob/dev/sdk/batch.py#L95)>). The old `wait` has been split into
-two separate boolean kwargs `wait` and `fetch_results`. - `wait` when set to `True` still makes the python statement blocking until the batch gets assigned a termination status (e.g. `DONE`, `ERROR`, `TIMED_OUT`) but doesn't trigger fetching of results. - `fetch_results` is a boolean which when set to `True` makes the python statement blocking until the batch has a termination status and then fetches the results for all the jobs of the batch.
+Reworked the `wait` logic
+when [creating a batch](https://github.com/pasqal-io/pasqal-cloud/blob/dev/sdk/__init__.py#L46)
+or [declaring it as complete](<(https://github.com/pasqal-io/pasqal-cloud/blob/dev/sdk/batch.py#L95)>). The old `wait`
+has been split into
+two separate boolean kwargs `wait` and `fetch_results`. - `wait` when set to `True` still makes the python statement
+blocking until the batch gets assigned a termination status (e.g. `DONE`, `ERROR`, `TIMED_OUT`) but doesn't trigger
+fetching of results. - `fetch_results` is a boolean which when set to `True` makes the python statement blocking until
+the batch has a termination status and then fetches the results for all the jobs of the batch.
 
-This enables the user to wait for the results and then implement its own custom logic to retrieve results (e.g. only fetch the results for the last job of the batch).
-This also fixes a bug where the user needed an extra call after the batch creation to the `get_batch` function to retrieve results. Now results will be properly populated after batch creation when setting `fetch_results=True`.
+This enables the user to wait for the results and then implement its own custom logic to retrieve results (e.g. only
+fetch the results for the last job of the batch).
+This also fixes a bug where the user needed an extra call after the batch creation to the `get_batch` function to
+retrieve results. Now results will be properly populated after batch creation when setting `fetch_results=True`.
 
 ### Fixed
 
@@ -343,7 +371,8 @@ This is the last released version before the implementation of the changelog.
 
 ### Added
 
-See commit history before [this commit](https://github.com/pasqal-io/pasqal-cloud/commit/7c703534f55012489550f7df116f3f326e741de5).
+See commit history
+before [this commit](https://github.com/pasqal-io/pasqal-cloud/commit/7c703534f55012489550f7df116f3f326e741de5).
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
-- Introduced a new filtering system using Pydantic classes for more flexible and robust filtering options
+- Introduced a new filtering system using classes for more flexible and robust filtering options
 - Added 'get_jobs' method, a feature to retrieve a list of jobs based on 'JobFilters'
 
 ### Breaking change
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to this project will be documented in this file.
 
-## [0.11.0] - 2024-06-14
+## [0.11.0] - 2024-06-27
 
 ### Added
 
--- a/README.md
+++ b/README.md
@@ -4,13 +4,15 @@ SDK to be used to access Pasqal Cloud Services.
 
 ## Installation
 
-To install the latest release of the `pasqal-cloud` (formerly pasqal-sdk), have Python 3.8.0 or higher installed, then use pip:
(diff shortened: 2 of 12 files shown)


pasqal_cloud/client.py
```diff
@@ -206,6 +209,51 @@ def rebatch(
         )["data"]
         return new_batch
 
+    def get_jobs(
```
nmenetrier, line 212: It looks like the other methods of this client are prefixed by underscores to discourage their direct usage by the users.
Maybe you should do the same ?

tests/test_batch.py
```diff
@@ -521,26 +421,15 @@ def test_rebatch_success(
...
-        # Check that correct url was requested with query params
+        # Converts 'jobs_ids' into 'id' as it is the field name used by the API
+        rebatch_filters = {
+            ("id" if k == "job_ids" else k): v for k, v in filters.items()
+        }
+
+        # Convert filters to the appropriate format for query parameters
+        query_params = build_query_params(rebatch_filters)
+
+        # Check that the correct url was requested with query params
```
nmenetrier, line 432: Thanks for the refactoring :slightly_smiling_face:

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.9.1] - 2024-05-16
```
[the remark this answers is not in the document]
Mildophin: Indeed, I forgot about it. I will update it after I get more reviews 👍🏻

pasqal_cloud/client.py
Mildophin: well I followed the same pattern as rebatch above, that is not a private method. I can change it and rebatch to a private method maybe ? To have something more consistent

CHANGELOG.md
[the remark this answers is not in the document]
Augustinio: I think since there's no backwards incompatible changes and we're still pre-1.0 it is only the patch that needs bumping

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
...
+
+### Added
```
Augustinio, line 7: In `### Changed` can you add the refactoring of the tests?

tests/test_batch.py
```diff
@@ -497,14 +405,6 @@ def test_get_batch_fetch_results_deprecated(
             {"status": ["DONE", "PENDING"]},
             {"min_runs": 10},
             {"max_runs": 10},
-            OrderedDict(
```
Augustinio, line 500: Why remove the test case that combines multiple filters together?

tests/test_job.py
```diff
@@ -17,3 +149,43 @@ def test_job_instantiation_with_extra_field(self, job):
...
+                "user_id": ["1", "2"],
+                "batch_id": [UUID(int=0x1), UUID(int=0x2)],
+                "status": [JobStatus.DONE, JobStatus.PENDING],
+                "min_runs": 10,
+                "max_runs": 20,
+                "errors": True,
+                "origin": [JobOrigin.AZURE, JobOrigin.CLIENT],
+                "start_date": datetime(2023, 1, 1),
+                "end_date": datetime(2023, 1, 1),
+            }
```
Augustinio, line 168: Here we have only one test case for all filters combined. Could they also be tested individually?

Augustinio (approved):
Just some minor comments about missing test cases. Are you also gonna add e2es?

tests/test_batch.py
```diff
@@ -521,26 +421,15 @@ def test_rebatch_success(
...
-                    value = value.isoformat()
-                query_params += f"{filter}={value}&"
-        # Replace : by %3A for datetimes
-        query_params = query_params.replace(":", "%3A")[:-1]
-
-        # Check that correct url was requested with query params
+        # Converts 'jobs_ids' into 'id' as it is the field name used by the API
+        rebatch_filters = {
+            ("id" if k == "job_ids" else k): v for k, v in filters.items()
+        }
```
MohamedAminMallek, line 427:
```suggestion
        if ids := filters.pop("job_ids", None):
            filters["id"] = ids
```
Maybe you can keep this part to avoid iterating over all attributes

tests/test_job.py
MohamedAminMallek, line 168: maybe add more test cases as you suggested in our live review session

pasqal_cloud/client.py
```diff
@@ -206,6 +209,51 @@ def rebatch(
...
+        project_id: Union[List[UUID], EmptyFilter],
+        user_id: Union[List[str], EmptyFilter],
+        batch_id: Union[List[UUID], EmptyFilter],
+        status: Union[Sequence[JobStatus], EmptyFilter],
+        min_runs: Union[int, EmptyFilter],
+        max_runs: Union[int, EmptyFilter],
+        errors: Union[bool, EmptyFilter],
+        origin: Union[Sequence[JobOrigin], EmptyFilter],
+        start_date: Union[datetime, EmptyFilter],
+        end_date: Union[datetime, EmptyFilter],
```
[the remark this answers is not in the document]
Augustinio: Yes I'm all for it!

tests/test_batch.py
Mildophin: I decided to keep it because it is one liner allowing to preserve the index of the key, okay it's an iteration for this use case the scope is alright I think, we do no risk any performance issue.

pasqal_cloud/utils/constants.py
```diff
@@ -0,0 +1,20 @@
+from enum import Enum
+
+
+class JobOrigin(Enum):
```
MatthieuMoreau0, line 4: There is no real reason for a user to filter on this so I don't think we need to expose it in the SDK. All batches created with the SDK should have the "CLIENT" origin

pasqal_cloud/__init__.py
```diff
@@ -251,28 +254,108 @@ def rebatch(
             job_ids: list of job ids to retry.
```
MatthieuMoreau0, line 254: We should clarify the behaviour of filters in this MR stating that they are multiplicative ('AND').

For each item give more details about how it behaves. 
For example il the user passes a list of statuses, we will rebatch all the jobs whose status is found in the list.

pasqal_cloud/__init__.py
```diff
@@ -251,28 +254,108 @@ def rebatch(
...
             status: status or list of statuses.
```
MatthieuMoreau0, line 255: According to the typehint this is always a list

pasqal_cloud/__init__.py
```diff
@@ -251,28 +254,108 @@ def rebatch(
...
+                batch_id=id,
+                filters=params,
             )
         except HTTPError as e:
             raise RebatchError(e) from e
         return Batch(**new_batch_data, _client=self._client)
 
+    def get_jobs(
+        self,
+        job_ids: Union[List[UUID], EmptyFilter] = EmptyFilter(),
```
MatthieuMoreau0, line 294: Let's be lenient on the user inputs and accept both UUID and str. We need to make sure to cast it to a known type (str or UUID) when building the `JobFilterParams`

pasqal_cloud/__init__.py
```diff
@@ -251,28 +254,108 @@ def rebatch(
...
+        if not isinstance(errors, EmptyFilter):
+            query_params["errors"] = errors
+        if not isinstance(origin, EmptyFilter):
+            query_params["origin"] = origin
+        if not isinstance(start_date, EmptyFilter):
+            query_params["start_date"] = start_date
+        if not isinstance(end_date, EmptyFilter):
+            query_params["end_date"] = end_date
+
+        params = JobFilterParams(**query_params)
```
MatthieuMoreau0, line 352: How about adding a dedicated method to `JobFilterParams` called `from_user_inputs` which gets rid of all the EmptyFilter (and do the casting if necessary) and returns the `JobFilterParams` object.

Benefits:
- less clutter in all the SDK methods
- centralize all the operations in the JobFilterParams. If we add a field it's going to be easier to remember to update that method
- Better type validation. Here there is no type validation when we do `JobFilterParams(**query_params)` because we go through a dictionnary. With a method we can type all the arguments.

MatthieuMoreau0 (changes requested):
The pagination is not supported in this MR, which is a critical aspect of the `get_jobs` method. This needs to be implemented before the MR can merge.

I suggested some small refactoring improvements as well.

pasqal_cloud/client.py
```diff
@@ -148,6 +150,43 @@ def _request(
...
+            params["offset"] = end
+            response: JSendPayload = self._request(
+                method=method, url=url, payload=payload, params=params
+            )
+
+            # Added to make sure the type is correctly set for mypy
+            # Should not happen
+            pagination_data = response.get("pagination")
+            if not pagination_data:
+                break
```
MohamedAminMallek, line 183: If for some reason this happens and it exits the loop, we will lose the data retrieved from the last request, so maybe extend the data to all_items before the `break`

pasqal_cloud/utils/filters.py
```diff
@@ -0,0 +1,122 @@
...
+        id: Filter by job ID(s), it can be a single UUID/str or a list of UUIDs/strs.
+        status: Filter by job status(es),
+        it can be a single JobStatus or a list of JobStatuses.
+        min_runs: Minimum number of runs.
+        max_runs: Maximum number of runs.
+        start_date: Filter by start date.
+        end_date: Filter by end date.
+    """
+
+    id: Optional[Union[List[Union[UUID, str]], Union[UUID, str]]] = None
```
MohamedAminMallek, line 54:
```suggestion
    job_ids: Optional[Union[List[Union[UUID, str]] = None
```

As discussed in our previous PP session, I prefer not to combine a list and a single value in the same attribute

MohamedAminMallek (approved):

Mildophin:
I realised I forgot to update the Readme, omw to update it.

Augustinio (approved):
LGTM, thank you for the refactoring of rebatch filters

Augustinio:
@MatthieuMoreau0 @Mildophin what if we added a mandatory batch_id param to get_jobs to ensure we only fetch jobs from a single batch, thus limiting the potential jobs fetched to 1000

pasqal_cloud/utils/filters.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class JobFilters(BaseFilters):
+    """
+    This class is used for providing filters to users for querying job-related data.
+
+    The filtering logic employs an AND logic between different fields,
+    meaning all conditions must be met simultaneously for a result to be included.
+
+    However, within fields that accept lists, each item in the list is treated as
```
MatthieuMoreau0, line 81: Suggestion:


```python
"""
Class to provide filters for querying jobs.

Setting a value for any attribute of this class will add a filter to the query.
When using several filters at the same time, the API will return elements who pass all filters at the same time:

    - If the filter attribute is a single element, the API will return jobs whose attribute match the provided value. 
    - If the filter attribute is a list, the API will return jobs whose value for this attribute is contained in that list.

Attributes:
<Fill in>
"""

And then you can add some examples, and for each example a full sentence to clarify what it does.

File: pasqal_cloud/utils/filters.py (hunk 15 above)
Comment (Mildophin) [reply pasqal_cloud/utils/filters.py:54]: I am in favor of giving the possibility to the user to choose as it is the most natural way for the users to filter what it needs. So I would keep it like this.

Then we can handle the data given and adapt it to the API if it only expects lists or single items.

File: README.md [hunk 17]
```diff
@@ -150,30 +158,73 @@ for job in batch.ordered_jobs:
     print(f"job-id: {job.id}, status: {job.status}, result: {job.result}")
 ```
 
+### Filter jobs
```
MatthieuMoreau0, line 161: Get a list of jobs

README.md
```diff
@@ -150,30 +158,73 @@ for job in batch.ordered_jobs:
...
+# Get the first 100 jobs, no filters applied
+sdk.get_jobs()
+
+# Get the first 40 jobs, no filters applied
+sdk.get_jobs(page_size=40)
+
+# Get the first 100 jobs from a given batch
+sdk.get_jobs(JobFilters(batch_id="batch_id"))
+
+# Get a list of specific jobs id
```
MatthieuMoreau0, line 182: Wording suggestion: Get two jobs using their IDs

README.md
```diff
@@ -150,30 +158,73 @@ for job in batch.ordered_jobs:
...
 ### Retry a batch of jobs
 
 It is possible to retry a selection of jobs from a CLOSED batch with the `rebatch` method.
 
 ```python
+from pasqal_cloud import RebatchFilters
+from pasqal_cloud import JobStatus
+
 # Retry all jobs from a given batch
 sdk.rebatch(batch.id)
```
MatthieuMoreau0, line 210: nit: batch.id => "batch_id"

pasqal_cloud/__init__.py
```diff
@@ -234,44 +236,115 @@ def cancel_batch(self, id: str) -> Batch:
...
         except HTTPError as e:
             raise RebatchError(e) from e
         return Batch(**new_batch_data, _client=self._client)
 
+    def get_jobs(
+        self,
+        filters: Optional[JobFilters] = None,
+        from_index: int = 0,
+        page_size: int = 100,
+    ) -> dict[str, Union[int, List[Job]]]:
```
MatthieuMoreau0, line 276: I'd recommend using either a TypedDict or a dedicated python class to hold the paginated data to benefit from type hinting

pasqal_cloud/__init__.py
```diff
@@ -234,44 +236,115 @@ def cancel_batch(self, id: str) -> Batch:
...
+            raise ValueError(
+                "Wrong type for 'page_size', it needs to be an integer, "
+                f"not a {type(filters)}"
+            )
+
+        if not (1 <= page_size <= 100):
+            raise ValueError(
+                "Wrong value for 'page_size', it has to be a positive integer "
+                "between 1 and 100"
+            )
```
MatthieuMoreau0, line 326: We will implement pagination for `get_batch` as well, which will have to perform the same checks. Let's introduce a class to regroup this validation logic

pasqal_cloud/__init__.py
```diff
@@ -234,44 +236,115 @@ def cancel_batch(self, id: str) -> Batch:
...
+            A dictionary with two keys:
+            - total: The total number of Jobs matching the filters.
+            - results: A list of Jobs matching the filters and
+                       pagination parameters provided.
+
+        Raises:
+            JobFetchingError: If fetching jobs call failed.
+        """
+        if not isinstance(from_index, int):
+            raise ValueError(
```
MatthieuMoreau0, line 306: This is a TypeError

pasqal_cloud/__init__.py
```diff
@@ -234,44 +236,115 @@ def cancel_batch(self, id: str) -> Batch:
...
+                f"not a {type(filters)}"
+            )
+
+        if from_index < 0:
+            raise ValueError(
+                "Wrong value for 'from_index', it has to be a positive integer"
+            )
+
+        if not isinstance(page_size, int):
+            raise ValueError(
```
MatthieuMoreau0, line 317: this is a TypeError

MatthieuMoreau0 (commented):
Looks good overall, some documentation suggestions

pasqal_cloud/client.py
```diff
@@ -176,37 +177,28 @@ def _cancel_batch(self, batch_id: str) -> Dict[str, Any]:
...
-    def rebatch(
-        self,
-        batch_id: str,
-        job_ids: Union[List[str], EmptyFilter],
-        status: Union[List[str], EmptyFilter],
-        min_runs: Union[int, EmptyFilter],
-        max_runs: Union[int, EmptyFilter],
-        start_date: Union[datetime, EmptyFilter],
-        end_date: Union[datetime, EmptyFilter],
+    def _rebatch(
```
oliver-gordon, line 180: This isn't a private method if it's used outside of this file.
Many of them aren't.

pasqal_cloud/__init__.py
```diff
@@ -234,44 +241,108 @@ def cancel_batch(self, id: str) -> Batch:
...
+
+        Returns:
+            PaginatedResponse: A class instance with two fields:
+                - total: An integer representing the total number of jobs
+                         matching the filters.
+                - results: A list of jobs matching the filters and pagination
+                           parameters provided.
+
+        Raises:
+            JobFetchingError: If fetching jobs call failed.
```
oliver-gordon, line 313: Missing ValueError exceptions

tests/test_job.py
```diff
@@ -1,7 +1,141 @@
...
+from tests.utils import build_query_params
 
 
 class TestJob:
+    @pytest.fixture(autouse=True)
+    @patch(
+        "pasqal_cloud.client.Auth0TokenProvider",
+        FakeAuth0AuthenticationSuccess,
+    )
+    def _init_sdk(self):
```
oliver-gordon, line 28: Doesn't need the `_` suffix

tests/utils.py
```diff
@@ -0,0 +1,36 @@
...
+    if isinstance(value, UUID):
+        return str(value)
+    if isinstance(value, (List, Tuple)):
+        return [convert_to_serializable(v) for v in value]
+    if isinstance(value, Dict):
+        return {k: convert_to_serializable(v) for k, v in value.items()}
+    return value
+
+
+def build_query_params(
```
oliver-gordon, line 20: Rather than just building the query parameters, why don't we actually encode the entire URL together with them? You're using an f-string above when you could just encode the entire URL. Less risk.

tests/test_job.py
```diff
@@ -1,7 +1,141 @@
...
+            "raw": ["1001", "1001", "0110", "1001", "0110"],
+        }
+
+    @pytest.mark.usefixtures("mock_request")
+    def test_get_job(self, job: Job):
+        """
+        Asserting the anticipated object is returned with
+        the mocked values when using SDK methods.
+        """
+        job_requested = self.sdk.get_job(job.id)
```
oliver-gordon, line 59: Sometimes wonder what's the point in tests like these, when we mock just to compare a returned, mocked string

pasqal_cloud/client.py
```diff
@@ -176,37 +177,28 @@ def _cancel_batch(self, batch_id: str) -> Dict[str, Any]:
...
-
         response: Dict[str, Any] = self._request(
             "POST",
             f"{self.endpoints.core}/api/v1/batches/{batch_id}/rebatch",
-            params=query_params,
+            params=filters.model_dump(exclude_unset=True),
         )["data"]
         return response
 
+    def _get_jobs(
```
oliver-gordon, line 190: Would argue this isn't private either, especially if we plan on exposing it to the `__init__.py` file

tests/test_job.py
Mildophin: I only applied the following rule of ruff: https://docs.astral.sh/ruff/rules/pytest-missing-fixture-name-underscore/

It raises an error during the pre-commit commit hook now as ruff is part of the dependencies. If you disagree with this rule I can change it.

pasqal_cloud/client.py
Mildophin: Indeed, however this is a whole different topic from the main subject of the PR but yes we should probably handle it differently. If you are fine with it I prefer to edit this in another PR more focused on the refactoring of the "strange parts" of this sdk 😄

tests/test_job.py
Mildophin: remainder from the past, but yes it does not bring much value

tests/utils.py
Mildophin: As discussed in private, it won't be implemented in this PR finally, but would be great to give up the use of f-strings to verify urls in our tests later, in another PR.

MatthieuMoreau0 (approved):

pasqal_cloud/utils/models.py
```diff
@@ -0,0 +1,150 @@
...
+    Class representing a paginated response structure.
+
+    Attributes:
+        total: The total number of data matching the query.
+        results: A list of data that match the query and pagination parameters
+                 provided.
+    """
+
+    total: int
+    results: List[Any]
```
MohamedAminMallek, line 150: I think it's useful to expose the offset to the user so it helps them iterate over pages easily
```suggestion
    offset: int
    results: List[Any]
```

PR #123, opened 2024-06-05
Get Jobs and Batch using v2 endpoints and download results from S3 via Presigned URLs

MohamedAminMallek:
### Checklist

#### Documentation

#### Tests

- [ ] Tests were run locally.


pasqal_cloud/client.py
```diff
@@ -149,6 +150,44 @@ def _request(
             "HTTP Client has encountered an issue it is unable to recover from."
         )
 
+    def _request_with_pagination(
```
MohamedAminMallek, line 153: I picked this method from @Mildophin PR but haven't copied the test yet

pasqal_cloud/client.py
```diff
@@ -160,10 +199,32 @@ def _send_batch(self, batch_data: Dict[str, Any]) -> Dict[str, Any]:
...
         return response
 
+    def _get_batch_jobs(self, batch_id: str) -> list[Dict[str, Any]]:
+        return self._request_with_pagination(
+            "GET",
+            f"{self.endpoints.core}/api/v2/jobs",
+            params={
+                "batch_id": batch_id,
+                "order_by": "ordered_id",
+                "order_by_direction": "ASC",
```
MohamedAminMallek, line 213: These order-by parameters are required to ensure that the jobs appear in the same order as they do in the GET /batches/{id} response. 

I am hesitant to set this as the default order-by value on core-api

pasqal_cloud/client.py
```diff
@@ -160,10 +199,32 @@ def _send_batch(self, batch_data: Dict[str, Any]) -> Dict[str, Any]:
...
+
+    def _get_job_results(self, job_id: str) -> Dict[str, Any]:
+        results_link = self._request(
+            "GET", f"{self.endpoints.core}/api/v1/jobs/{job_id}/results_link"
+        )["data"]["results_link"]
+        if results_link:
+            try:
+                return self._request("GET", results_link)
+            except HTTPError:
+                pass
```
MohamedAminMallek, line 225: Avoid raising an exception if the pre-signed URL points to a non-existent file (result not uploaded yet)

I noticed in the e2e tests that some jobs move to the DONE status while the results are still not yet uploaded

pasqal_cloud/client.py
MatthieuMoreau0: I'll make the same suggestion to rename this `request_all_pages` then :wink:

tests/test_batch.py
```diff
@@ -346,7 +347,7 @@ def test_get_job_error(
         assert (
             mock_request_exception.last_request.url
             == f"{self.sdk._client.endpoints.core}"
-            f"/api/v1/jobs/{job.id}"
+            f"/api/v2/jobs/{job.id}"
```
oliver-gordon, line 350: So are we considering V1 deprecated now? Do we no longer want to support testing its behaviors?

tests/conftest.py
```diff
@@ -42,6 +43,15 @@ def mock_core_response(request):
         return result
 
 
+def mock_s3_presigned_url_response(request):
```
oliver-gordon, line 46: docstr and types please

pasqal_cloud/client.py
MatthieuMoreau0: What's the current default ordering? If this can help keeping jobs in the same order, I'm all for it

tests/test_batch.py
MatthieuMoreau0: The v1 endpoints are going to be tested in our e2es which target older versions of the SDK.

The responsibility of the unit test is only to test the features of the SDK, which means only the endpoint the SDK uses. That's my understanding at least.

On a tangent note, I think the v1 endpoint is indeed considered deprecated now

pasqal_cloud/client.py
Mildophin: It is, from what I see:

order_by= "created_at"
order_by_direction: "DESC"

MohamedAminMallek: The abstract repository class applies default ordering by `id` and `created_at`. Since this ordering is used for all endpoints, I don't think it's necessary to change it specifically for jobs. It's better to maintain consistent default ordering across all endpoints. We can still change the ordering when calling the endpoint, as done here.

tests/test_batch.py
oliver-gordon: Not sure why that would be a tangent note, it was the first question I wanted an answer for.

MatthieuMoreau0: Yeah I rewrote that paragraph and messed up oops. 

Anyway the point was I don't think v1 endpoints should be tested here, do you disagree?

pasqal_cloud/batch.py
```diff
@@ -230,5 +227,6 @@ def refresh(self) -> None:
     def _update_from_api_response(self, data: Dict[str, Any]) -> None:
         """Update the instance in place with the response body of the batch API"""
         updated_batch = Batch(**data, _client=self._client)
-        for field, value in updated_batch:
+        for field in [*list(updated_batch.model_fields), "_ordered_jobs"]:
```
MatthieuMoreau0, line 230: this is hacky, can you add a comment about it?

pasqal_cloud/client.py
```diff
@@ -149,6 +150,44 @@ def _request(
             "HTTP Client has encountered an issue it is unable to recover from."
         )
 
+    def _request_all_pages(
+        self,
+        method: str,
+        url: str,
+        payload: Optional[Union[Mapping, Sequence[Mapping]]] = None,
+        params: Optional[Dict[str, Any]] = None,
+    ) -> List[Dict[str, Any]]:
```
MatthieuMoreau0, line 159: docstring please

pasqal_cloud/client.py
```diff
@@ -149,6 +150,44 @@ def _request(
...
+        total_nb_items = pagination_data["total"]
+        end = pagination_data["end"]
+        while end < total_nb_items:
+            params["offset"] = end
+            response: JSendPayload = self._request(
+                method=method, url=url, payload=payload, params=params
+            )
+
+            # Added to make sure the type is correctly set for mypy
+            # Should not happen
```
MatthieuMoreau0, line 180: I don't think this comment is at the right line, is it?

pasqal_cloud/client.py
MatthieuMoreau0: >These order-by parameters are required to ensure that the jobs appear in the same order as they do in the GET /batches/{id} response.

Can you add this as a comment in the code so that we know why we define an explicit ordering

pasqal_cloud/client.py
```diff
@@ -160,10 +199,32 @@ def _send_batch(self, batch_data: Dict[str, Any]) -> Dict[str, Any]:
...
+            "GET",
+            f"{self.endpoints.core}/api/v2/jobs",
+            params={
+                "batch_id": batch_id,
+                "order_by": "ordered_id",
+                "order_by_direction": "ASC",
+            },
+        )
+
+    def _get_job_results(self, job_id: str) -> Any:
```
MatthieuMoreau0, line 217: Can we restrict the return type?

pasqal_cloud/job.py
```diff
@@ -7,6 +7,11 @@
 from pasqal_cloud.errors import JobCancellingError
 
 
+class ResultType(TypedDict):
```
MatthieuMoreau0, line 10: How does this behave if we add new attributes to result later? Will this break this version of the SDK?

MatthieuMoreau0, line 10: Why is this suffixed by "Type"?

MatthieuMoreau0 (commented):
Mostly doc and type hinting nitpicks. I have one question about ResultType behaviour

pasqal_cloud/client.py
MohamedAminMallek: In the current method, we are using `_request` to download results from S3. However, `_request` is using JSendPayload as a return typehint, which is not correct when we are not requesting core-api. That's why I used Any. 
But we can directly use `requests` in `_get_job_results` to have something like this:

 ```
   def _get_job_results(self, job_id: str) -> Optional[JobResult]:
        results_link = self._request(
            "GET", f"{self.endpoints.core}/api/v1/jobs/{job_id}/results_link"
        )["data"]["results_link"]
        if results_link:
            try:
                response = requests.get(results_link)
                response.raise_for_status()
                data = response.json()
                return JobResult(
                    raw=data.pop("raw", None), counter=data.pop("counter", None), **data
                )
            except HTTPError:
                pass
        return None
```
And we can, of course, have another internal method that performs HTTP calls, which will be used by `_request` and `_get_job_results`

what do you think?

File: pasqal_cloud/job.py (hunk 10 above)
Comment (MohamedAminMallek) [reply pasqal_cloud/job.py:10]: -> `JobResult` ✅

File: pasqal_cloud/client.py (hunk 9 above)
Comment (MatthieuMoreau0) [reply pasqal_cloud/client.py:217]: ok for me. I don't think we should use our `_request` method since it is meant to call our APIs - it automatically injects auth in the headers and so on which we don't want to interact with s3

File: pasqal_cloud/batch.py [hunk 11]
```diff
@@ -230,5 +227,10 @@ def refresh(self) -> None:
     def _update_from_api_response(self, data: Dict[str, Any]) -> None:
         """Update the instance in place with the response body of the batch API"""
         updated_batch = Batch(**data, _client=self._client)
-        for field, value in updated_batch:
+        # Private fields are not exposed by the model_fields method, so we need to
+        # explicitly add _ordered_jobs to ensure it is compied from the updated_batch
```
MatthieuMoreau0, line 231: typo: compied

MatthieuMoreau0 (approved):

Mildophin (approved):
Good to me as most of the comments suggestions were applied 👍🏻

PR #124, opened 2024-06-07
[DRAFT] Expose add_jobs to __init__ for use in client

oliver-gordon:
### Description

We need a method of submitting Jobs to a preexisting batch exposed on the interface that's being used by pulser-pasqal. I've also included a required method to mark a batch as complete via the SDK interface.

I've also updated some small nits that weren't quite right. We've got method prefixes for methods that are clearly exported into other files.

### Checklist

#### Documentation

#### Tests

- [ ] Tests were run locally.

### DIFF
--- a/pasqal_cloud/__init__.py
+++ b/pasqal_cloud/__init__.py
@@ -32,7 +32,7 @@ from pasqal_cloud.errors import (
     BatchCancellingError,
     BatchCreationError,
     BatchFetchingError,
-    BatchAlreadyCompleteError,
+    BatchSetCompleteError,
     DeviceSpecsFetchingError,
     JobCancellingError,
     JobCreationError,
@@ -301,12 +301,12 @@ class SDK:
                 job_rsp = self._get_job(id)
         return Job(**job_rsp, _client=self._client)
 
-    def add_jobs(self, batch_id: str, jobs: List[CreateJob]):
+    def add_jobs(self, batch_id: str, jobs: List[CreateJob]) -> Batch:
         """
         Add jobs to an already existing batch.
 
         Args:
-            batch_ID: A unique identifier for the batch
+            batch_ID: A unique identifier for the batch data.
             Jobs: a collection of CreateJob payloads
 
         Returns:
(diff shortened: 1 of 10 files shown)


oliver-gordon:
I'll update the release version before this MR goes in because there are several MRs that could go in at any point right now and I don't want to push to update multiple times.

pasqal_cloud/__init__.py
```diff
@@ -254,7 +256,8 @@ def rebatch(
             mas_runs: retry jobs with less or as many as this amount of runs.
             start_date: retry jobs created at or after this datetime.
             end_date: retry jobs created at or before this datetime.
-
+        Returns:
+            An instance of a Batch model
```
Mildophin, line 260: What about "The instance of the re-created batch model" instead to make it more precise ?

Mildophin (approved):
Good to me 👍🏻

pasqal_cloud/__init__.py
oliver-gordon: I'll add some more specificity

pasqal_cloud/__init__.py
```diff
@@ -256,8 +256,12 @@ def rebatch(
             mas_runs: retry jobs with less or as many as this amount of runs.
             start_date: retry jobs created at or after this datetime.
             end_date: retry jobs created at or before this datetime.
+
         Returns:
-            An instance of a Batch model
+            An instance of a rescheduled Batch model. The fields
+            can differ from the original batch as the record
+            is recreated as to prevent modfifying the original batch.
```
Mildophin, line 263: There is a typo in case you missed it.

Thx for the extra documentation it is clearer now 😄

pasqal_cloud/__init__.py
```diff
@@ -298,22 +305,68 @@ def get_job(self, id: str, wait: bool = False) -> Job:
                 job_rsp = self._get_job(id)
         return Job(**job_rsp, _client=self._client)
 
+    def add_jobs(self, batch_id: str, jobs: List[CreateJob]) -> Batch:
```
MatthieuMoreau0, line 308: These two methods are duplicates of methods from the `batch` class; I understand we want to expose them at the top level for easy access in pasqal cloud but this makes me wonder a few things:
- a little duplication is fine but maybe we could manage to use these new functions in the Batch method to avoid the the duplication of the error catching
- Should we ditch entirely the batch class methods and make it a simple dataclass with no methods? I don't see much benefits for them at this point. For backwards compat, we probably have to stick with them 
- We really need to rename on the backend side the "complete" flag to "open"

MatthieuMoreau0 (approved):
Looks good, I left a couple non-blocking thoughts for the Batch methods.

Don't forget to update the _version file to release this when merged and the changelog. Please drop the DRAFT prefix in the title as well

pasqal_cloud/__init__.py
oliver-gordon: Given the terminology we use absolutely everywhere is 'open batch', changing the references from `complete` to `open` makes sense. Otherwise we should stop calling them open batches and rather noncomplete batches. I'd like it if we can avoid more rebatch-esque naming scenarios.

It does not look like we're using the methods bound to the Batch model, so I'd be fine with removing them.

Regarding to backwards compatibility, we can leave things for now but would be nice to know we're able to depreciate things eventually. Not much benefit providing the same interface in two places because someone doesn't want to update their script.

PR #127, opened 2024-06-20
[IMPROVEMENT] Add a retry mechanism when downloading job results

MohamedAminMallek:
### Description

- Add a retry decorator
- Use the decorator to enable retry when downloading job results
- Refactor _request to use the new decorator

### Checklist

#### Documentation

#### Tests

- [x] Tests were run locally.

### DIFF
--- a/pasqal_cloud/client.py
+++ b/pasqal_cloud/client.py
@@ -105,7 +105,7 @@ class Client:
     @retry_http_error(
         max_retries=5, retry_status_code={408, 425, 429, 500, 502, 503, 504}
     )
-    def _request(
+    def _authenticated_request(
         self,
         method: str,
         url: str,
@@ -148,7 +148,7 @@ class Client:
 
         if not params:
             params = {}
-        first_page_response = self._request(
+        first_page_response = self._authenticated_request(
             method=method, url=url, payload=payload, params=params
         )
         all_items: List[Dict] = first_page_response["data"]
@@ -161,7 +161,7 @@ class Client:
         end = pagination_data["end"]
         while end < total_nb_items:
             params["offset"] = end
-            response: JSendPayload = self._request(
+            response: JSendPayload = self._authenticated_request(
                 method=method, url=url, payload=payload, params=params
             )
 
@@ -179,7 +179,7 @@ class Client:
 
     def send_batch(self, batch_data: Dict[str, Any]) -> Dict[str, Any]:
         batch_data.update({"project_id": self.project_id})
-        response: Dict[str, Any] = self._request(
+        response: Dict[str, Any] = self._authenticated_request(
             "POST",
             f"{self.endpoints.core}/api/v1/batches",
             batch_data,
@@ -187,7 +187,7 @@ class Client:
         return response
 
     def get_batch(self, batch_id: str) -> Dict[str, Any]:
-        response: Dict[str, Any] = self._request(
+        response: Dict[str, Any] = self._authenticated_request(
             "GET", f"{self.endpoints.core}/api/v2/batches/{batch_id}"
         )["data"]
         return response
@@ -221,7 +221,7 @@ class Client:
         This function request a presigned-url for a specific job_id from the core API.
         Once the presigned-url is obtained, it downloads the result from S3.
         """
-        results_link = self._request(
+        results_link = self._authenticated_request(
             "GET", f"{self.endpoints.core}/api/v1/jobs/{job_id}/results_link"
         )["data"]["results_link"]
         if results_link:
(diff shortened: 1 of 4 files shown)


pasqal_cloud/client.py
```diff
@@ -230,18 +206,22 @@ def get_batch_jobs(self, batch_id: str) -> list[Dict[str, Any]]:
...
             try:
-                response = requests.get(results_link)
-                response.raise_for_status()
-                data = response.json()
-                return JobResult(
-                    raw=data.pop("raw", None), counter=data.pop("counter", None), **data
-                )
+                return self._download_results(results_link)
             except HTTPError:
                 pass
```
MohamedAminMallek, line 226: In case of an HTTPError even after the retry, do you think we should raise the exception or just set the results to None?

The reason I chose to return None is that we currently set the job status to Done before uploading the result, so we may have a pre-signed URL but no results uploaded yet.

What do you think?

pasqal_cloud/utils/retry.py
```diff
@@ -0,0 +1,60 @@
...
+)
+
+from requests import HTTPError
+
+Param = ParamSpec("Param")
+RT = TypeVar("RT")  # return type
+
+
+def retry_http_error(
+    max_retries: int = 5, retry_status_code: Optional[Iterable[int]] = None
```
oliver-gordon, line 18: If possible are we able to begin typing Optional with  `N | None`

tests/test_client.py
```diff
@@ -223,6 +223,17 @@ def _mock_sleep(self):
         with patch("time.sleep"):
             yield
 
+    def test_download_results_retry_on_exception(
```
oliver-gordon, line 226: doc for the test

pasqal_cloud/client.py
```diff
@@ -230,18 +206,22 @@ def get_batch_jobs(self, batch_id: str) -> list[Dict[str, Any]]:
...
+    @retry_http_error(max_retries=2)
+    def _download_results(self, results_link: str) -> JobResult:
+        response = requests.get(results_link)
+        response.raise_for_status()
+        data = response.json()
+        return JobResult(
+            raw=data.pop("raw", None), counter=data.pop("counter", None), **data
+        )
+
     def get_job_results(self, job_id: str) -> Optional[JobResult]:
```
oliver-gordon, line 218: None of these funcs public or priv have any docstr just to account for any necessary details

pasqal_cloud/client.py
```diff
@@ -230,18 +206,22 @@ def get_batch_jobs(self, batch_id: str) -> list[Dict[str, Any]]:
             # the same order as they do in the GET /batches/{id} response
         )
 
+    @retry_http_error(max_retries=2)
+    def _download_results(self, results_link: str) -> JobResult:
+        response = requests.get(results_link)
```
oliver-gordon, line 211: As much as I don't care what interface you're using for your HTTP request. We are now splitting our use between the internal `_request(method_type: str)` and `requests.method_type()` so maybe we should pick one or the other for consistency

MatthieuMoreau0: From my understanding `_request` is used to speak to our APIs and automatically attaches authentication headers to the request; we don't want these when fetching the results in s3, which is why we use `request` directly here.

We should maybe clarify this by renaming the `_request` method to `_authenticated_request` or something. We could also add in the docstring that this method is meant to be used when calling our APIs.

Or we could make the `_request` function more generic to make auth headers optional.

I have a preference for the first option personally.

pasqal_cloud/utils/retry.py
MohamedAminMallek: N | None syntax is only supported in 3.10

pasqal_cloud/client.py
MohamedAminMallek: Do we want to rename it in this PR? If so, I will do it as the last commit before merging the PR to facilitate review

MatthieuMoreau0: Yeah since this MR introduces another method to make http request, I think it should take care of renaming to keep things clear

pasqal_cloud/client.py
oliver-gordon: If we get HTTP errors after N retries why would we not raise the exception? Technically everything has gone wrong. The retries are the mechanism to try subdue bubbling exceptions but it would come off a tad unreliable if we return None to mask exceptions, no?

oliver-gordon (approved):

MatthieuMoreau0 (approved):

Mildophin (approved):

PR #128, opened 2024-07-03
[IMPROV] Extend the retry mechanism to handle runtime exceptions

MohamedAminMallek:
### Description

Our internal e2e tests are failing due to random network issues with S3. Therefore, we have decided to extend the retry mechanism to handle exceptions, rather than relying only on HTTP status codes

### Checklist

#### Documentation

#### Tests

- [ ] Tests were run locally.

### DIFF
--- a/pasqal_cloud/client.py
+++ b/pasqal_cloud/client.py
@@ -206,7 +206,7 @@ class Client:
             # the same order as they do in the GET /batches/{id} response
         )
 
-    @retry_http_error(max_retries=2, retry_exceptions=(requests.ConnectionError,))
+    @retry_http_error(max_retries=5, retry_exceptions=(requests.ConnectionError,))
     def _download_results(self, results_link: str) -> JobResult:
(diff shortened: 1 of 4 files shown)


pasqal_cloud/client.py
```diff
@@ -206,7 +206,7 @@ def get_batch_jobs(self, batch_id: str) -> list[Dict[str, Any]]:
             # the same order as they do in the GET /batches/{id} response
         )
 
-    @retry_http_error(max_retries=2)
+    @retry_http_error(max_retries=2, retry_exceptions=(requests.ConnectionError,))
```
oliver-gordon, line 209: So they are failing due to network issues or synchronicity issues? I thought the test was running faster than an updated resource and this was the failure. We _know_ it's due to a network issue?

Should we bump it to 5? If something fails due to a network issues, it could well fail again immediately if the expected first delay is 1 second

PR #129, opened 2024-07-10
[FEAT] Add cancel_jobs method

Mildophin:
### Description

This PR add the ability to cancel a group of jobs from a specific batch, it takes in consideration the following filtering options:
 - job id
 - min_runs
 - max_runs
 - start_date
 - end_date
 
 It follows the same pattern of filtering I introduced for getting a list of jobs in [this PR](https://github.com/pasqal-io/pasqal-cloud/pull/121).

#### Remaining Tasks

E2E tests

### Checklist

#### Documentation

#### Tests

- [x] Tests were run locally.


pasqal_cloud/__init__.py
```diff
@@ -435,6 +436,47 @@ def cancel_job(self, id: str) -> Job:
 
         return Job(**job_rsp, _client=self._client)
 
+    def cancel_jobs(
+        self,
+        batch_id: Union[UUID, str],
+        filters: Optional[CancelJobFilters] = None,
```
oliver-gordon, line 442: Why not just default `filters` to `= CancelJobFilters()`? Then you can remove some of the logic below

pasqal_cloud/utils/filters.py
```diff
@@ -76,6 +76,46 @@ def status_enum_to_string(self, job_statuses: List[JobStatus]) -> List[str]:
...
+        max_runs: Maximum number of runs.
+        start_date: Retry jobs created at or after this datetime.
+        end_date: Retry jobs created at or before this datetime.
+    """
+
+    id: Optional[Union[List[Union[UUID, str]], Union[UUID, str]]] = None
+    min_runs: Optional[int] = None
+    max_runs: Optional[int] = None
+    start_date: Optional[datetime] = None
+    end_date: Optional[datetime] = None
```
MohamedAminMallek, line 104: `RebatchFilters`, `CancelJobFilters`, and `SuspendJobFilters` share some filter attributes. Can we move these shared attributes to `BaseFilters`?

pasqal_cloud/__init__.py
Mildophin: Because i think it is easier to understand it is optional like this.

Correct me if i am wrong but if I use this instead, I still need to make sure it is an instance of a Filter class passed right ?

pasqal_cloud/utils/filters.py
Mildophin: Indeed it is possible, I will merge it

MohamedAminMallek (approved):

PR #131, opened 2024-07-31
[Refactor] Use open instead of complete

Augustinio:
### Description

- Introduced `open` to replace `complete` argument
- Refactor fixture data and tests

### Checklist

#### Documentation

#### Tests

- [x] Tests were run locally.


pasqal_cloud/batch.py
```diff
@@ -181,23 +183,23 @@ def retry(self, job: Job, wait: bool = False) -> None:
         except JobCreationError as e:
             raise JobRetryError from e
 
-    def declare_complete(self, wait: bool = False, fetch_results: bool = False) -> None:
-        """Declare to PCS that the batch is complete and returns an updated
+    def close(self, wait: bool = False, fetch_results: bool = False) -> None:
```
MohamedAminMallek, line 186: breaking change ?

pasqal_cloud/client.py
```diff
@@ -234,7 +234,7 @@ def get_job_results(self, job_id: str) -> Optional[JobResult]:
             return self._download_results(results_link)
         return None
 
-    def complete_batch(self, batch_id: str) -> Dict[str, Any]:
+    def close_batch(self, batch_id: str) -> Dict[str, Any]:
```
MohamedAminMallek, line 237: breaking change ?

pasqal_cloud/__init__.py
```diff
@@ -390,7 +404,6 @@ def add_jobs(self, batch_id: str, jobs: List[CreateJob]) -> Batch:
 
         Raises:
             JobCreationError, which spawns from a HTTPError.
-            BatchAlreadyCompleteError.
```
Mildophin, line 393: should raise OnlyCompleteOrOpenCanBeSet instead, could you add it ?

pasqal_cloud/__init__.py
```diff
@@ -161,6 +163,18 @@ def create_batch(
             BatchCreationError: If batch creation failed
             BatchFetchingError: If batch fetching failed
         """
+        if complete is not None and open is not None:
+            raise OnlyCompleteOrOpenCanBeSet
```
oliver-gordon, line 167: I understand the want for the exception but could we simply not ignore `complete` if `open` is set? while still outputting the warning since it is a deprecated field

pasqal_cloud/__init__.py
```diff
@@ -127,7 +128,8 @@ def create_batch(
         self,
         serialized_sequence: str,
         jobs: List[CreateJob],
-        complete: bool = True,
+        complete: Optional[bool] = None,
+        open: Optional[bool] = None,
```
oliver-gordon, line 132: Or even if we retrained the default value, because we set it to True as the parameter before, meaning it would need to be False and then we just raise a warning

pasqal_cloud/batch.py
Augustinio: I didn't notice that @MatthieuMoreau0 should I refrain from renaming these methods since it's a breaking change?
If I understand correctly we're keeping complete and open together so that users don't have to update their scripts. This would probably defeat that purpose.

MatthieuMoreau0: imo we should:
- add a `close` method like you did here
- keep the `declare_complete` method
  - make it raise a deprecation warning for the user to invite them to use `close` instead
  - use the `close` method in its body so that we do not duplicate the logic

pasqal_cloud/__init__.py
Augustinio: The ticket specified the following:
```
n the SDK we should accept both a complete and open kwarg. If the user sets both, raise an error. Convert these kwargs into the "open" attribute for the batch creation request.
```

pasqal_cloud/__init__.py
Augustinio: I think you're mistaken. This is the add jobs method.

pasqal_cloud/__init__.py
Augustinio: As the current behaviour in the POST batch endpoint is to raise an exception as per the specifications of the tickets and it has already been released to prod I would rather keep this one consistent and raise the exception rather than ignoring complete.

pasqal_cloud/__init__.py
Mildophin: Indeed, mb

pasqal_cloud/__init__.py
MatthieuMoreau0: The ticket specs can be challenged if we find a better solution! I like the suggestion as well, it makes the interface a bit more lenient on the user inputs

pasqal_cloud/batch.py
```diff
@@ -28,7 +28,8 @@ class Batch(BaseModel):
     device until all its jobs are done and declared complete.
 
     Attributes:
-        - complete: Whether the batch has been declared as complete.
+        - open: Whether the batch accepts more jobs or not.
+        - complete: Opposite of open.
```
MatthieuMoreau0, line 32: mark it as deprecated

MatthieuMoreau0 (approved):
No other main comments than the other reviewers. I made a few documentation suggestions.

Don't forget to keep support for the `complete_batch` method for backwards compatibility. It looks like it has been removed in the current implementation.

pasqal_cloud/__init__.py
Augustinio: It would require an MR on public-apis side to make for a consistent behaviour which I don't think this is worth but if there's a consensus I'll abide

Augustinio: As a last check that this is what we want. Because I believe we're trading leniency for clarity:
The args on the batch would look like:

```
...
open: bool = False
complete: bool = True
```

Then the following conditions would lead to the following states.

- open: True , complete: True -> the batch is open, complete is ignored
- open: False (default), complete: True (default) -> The batch is closed
- open: True , complete: False - > the batch is closed, complete is ignored
- open: False (default), complete: False -> The batch is opened, (here open is ignored)


@MohamedAminMallek @oliver-gordon @MatthieuMoreau0 let me know if I got this right

MatthieuMoreau0: Thanks for pointing this out. 

On second thought I indeed prefer raising an exception. For old scripts running on the latest SDK version there will be no error since it will only specify `complete` - this is good for backwards compatibility's sake. If the user makes an effort to use the new `open` argument we can expect them to get rid of the old `complete` one.

That should also simplify the logic

Mildophin: I am in favor of raising an exception as well.

Mildophin (approved):
Good to me overall.

PR #132, opened 2024-08-02
[CLEANUP] Remove deprecated field 'group' and deprecated import 'from sdk'

Mildophin:
### Description

Simple cleanup PR that removed the deprecated field 'group', making compulsory the use of 'project' instead.

Also it removes the old way of importing this sdk with 'from sdk', it's been deprecated for more than year now and it is removed in this PR. We have to use 'from pasqal_cloud' from now.

### Checklist

#### Documentation

#### Tests

- [ ] Tests were run locally.


Augustinio (approved):
LGTM

Augustinio (approved):

pasqal_cloud/__init__.py
```diff
@@ -59,14 +59,13 @@ class SDK:
 
     def __init__(
         self,
+        project_id: str,
         username: Optional[str] = None,
         password: Optional[str] = None,
         token_provider: Optional[TokenProvider] = None,
         endpoints: Optional[Endpoints] = None,
         auth0: Optional[Auth0Conf] = None,
         webhook: Optional[str] = None,
```
awennersteen, line 68: mention this in the changelog since a common way before of instantiating the sdk would have been
```python
sdk = SDK("awennersteen", "super-secret", project_id="my-project_id")
```
And this won't work anymore.

Mildophin: Indeed, bad python practices do not help haha, I will update the changelog for this potential breaking change.

MohamedAminMallek: I'm not sure I understand why you decided to move the project_id to the first argument

Mildophin: -> non-default parameter follows default parameter

Python things if I correctly understood how it works

MatthieuMoreau0: It's probably not worth the risk of breaking scripts of user to make it non-optional

MatthieuMoreau0 (approved):

pasqal_cloud/__init__.py
oliver-gordon: If this is going to cause significant breaks, we could accommodate by still defaulting to `None` and keep on raising an exception in the scenario it's missing.

We have optional input for several fields that are very obviously not optional fields. Can't authenticate without a password.

Mildophin: Then I will put it back as optional and add a warning if not given. Thx for the feedback.

MatthieuMoreau0: You should raise an exception not only a warning!

Mildophin: I meant, an exception for sure, but also a warning saying that project_id will be the first argument later, like a deprecation warning.

oliver-gordon: I'd avoid the warning, we should avoid preemptive statements because nothings designed or planned.

What if we simply change our mind on something?

CHANGELOG.md
```diff
@@ -2,14 +2,24 @@
 
 All notable changes to this project will be documented in this file.
 
-## [0.11.4] - 2024-07-31
+## [0.12.0] - 2024-08-29
+
+### Breaking change
+
+- 'from pasqal_cloud' import has completely replaced the deprecated import 'from sdk'
+- 'project' field has completely replaced the deprecated field 'group'
```
oliver-gordon, line 10: field is `project_id` and this MR is simply removing `group_id`, We introduced project_id a while ago

CHANGELOG.md
```diff
@@ -2,14 +2,24 @@
...
+- Now you need to declare the 'project_id' as the first argument when instantiating an SDK class
```
oliver-gordon, line 11: outdated

oliver-gordon (approved):

CHANGELOG.md
Mildophin: Should I write "'group' field is now removed, use 'project' field instead" ? Not sure how to make the message clear when we read the sentence.

CHANGELOG.md
Mildophin: good spot

PR #135, opened 2024-09-04
[FEAT] Switch to V2 endpoints for some endpoints (#1281)

Mildophin:
### Description

Switch to V2 endpoints for the following endpoints:
- PUT cancel jobs
- POST add jobs
- PUT cancel batch
- PUT complete batch
- PUT cancel job

The V2 remove the 'results' field in the response, field present in the V1 of these endpoints.

### Checklist

#### Documentation

#### Tests

- [ ] Tests were run locally.


Augustinio:
Technically if we remove a key from the response, doesn't it constitute a breaking change? If so the minor should be bumped.

Mildophin:
> Technically if we remove a key from the response, doesn't it constitute a breaking change? If so the minor should be bumped.

Indeed it is, just realised it, I will bump it to 0.13.0

Augustinio (approved):

nmenetrier (approved):

MohamedAminMallek (approved):

CHANGELOG.md
```diff
@@ -2,25 +2,36 @@
...
+## [0.13.0] - 2024-09-04
+
+### Breaking change
+
+- The following methods now use the V2 that do not return the results any more:
+    - Cancel a batch
+    - Cancel a job
+    - Cancel a group of jobs
+    - Add jobs to a batch
+    - Close a batch
```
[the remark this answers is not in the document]
Mildophin: Amin is actually right as we have a property that set the result in the job Class, so the user will not notice any difference, I bring back it to 0.12.1 and a different changelog.

CHANGELOG.md
```diff
@@ -2,25 +2,36 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.12.1] - 2024-09-10
+
+### Changed
+
+- Better performances for the following methods:
```
oliver-gordon, line 9: Nit: There is no performance gain in this SDK. You have just updated the endpoints to v2.

oliver-gordon (approved):

CHANGELOG.md
Mildophin: However the user does not see it, it does 0 difference for it, should I write it down instead ?

PR #138, opened 2024-09-19
Allow unauthenticated users to access device specifications

MohamedAminMallek:
### Checklist

#### Documentation

#### Tests

- [ ] Tests were run locally.

### DIFF
--- a/pasqal_cloud/client.py
+++ b/pasqal_cloud/client.py
@@ -44,33 +44,45 @@ class EmptyFilter:
 
 
 class Client:
-    authenticator: AuthBase
-
     def __init__(
         self,
-        project_id: str,
+        project_id: str | None = None,
         username: Optional[str] = None,
         password: Optional[str] = None,
         token_provider: Optional[TokenProvider] = None,
         endpoints: Optional[Endpoints] = None,
         auth0: Optional[Auth0Conf] = None,
     ):
-        if not username and not token_provider:
-            raise ValueError(
-                "At least a username or TokenProvider object should be provided."
-            )
+        self.endpoints = self._make_endpoints(endpoints)
+        self._project_id = project_id
+        self.user_agent = f"PasqalCloudSDK/{sdk_version}"
+
         if token_provider is not None:
+            self.project_id
             self._check_token_provider(token_provider)
 
-        self.endpoints = self._make_endpoints(endpoints)
-
         if username:
+            self.project_id
             auth0 = self._make_auth0(auth0)
             token_provider = self._credential_login(username, password, auth0)
 
-        self.authenticator = HTTPBearerAuthenticator(token_provider)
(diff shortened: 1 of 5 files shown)


MatthieuMoreau0 (commented):
The feature looks good. I made some refactoring suggestions.

pasqal_cloud/client.py
```diff
@@ -44,33 +44,41 @@ class EmptyFilter:
...
 
 class Client:
-    authenticator: AuthBase
+    authenticator: AuthBase | None
 
     def __init__(
         self,
-        project_id: str,
+        project_id: str | None = None,
         username: Optional[str] = None,
```
Mildophin, line 52: authentificator: Union[str, None]

project_id: Optional[str] = None,

what about this to be consistent with the others ? The pipe was added in python 3.10 however we need to make sure pasqal-cloud is compatible with python 3.8

pasqal_cloud/client.py
```diff
@@ -120,6 +128,12 @@ def _authenticated_request(
         payload: Optional[Union[Mapping, Sequence[Mapping]]] = None,
         params: Optional[Mapping[str, Any]] = None,
     ) -> JSendPayload:
+        if self.authenticator is None:
+            raise ValueError(
+                "Authentication required. Please provide your credentials when"
+                " initiating the client."
```
MatthieuMoreau0, line 134: initiating => initializing

MatthieuMoreau0 (approved):
Thanks for the changes, looks good!

pasqal_cloud/client.py
MohamedAminMallek: In a previous PR, @oliver-gordon  suggested using the new way to declare optional parameters (`type | None = None`). I think, this is the right approach, especially since it’s the new standard and our checks/tests are passing

pasqal_cloud/client.py
```diff
@@ -325,7 +339,17 @@ def cancel_workload(self, workload_id: str) -> Dict[str, Any]:
...
+            )["data"]
+            return response
+        return self.get_public_device_specs()
+
+    def get_public_device_specs(self) -> Dict[str, str]:
+        response = requests.request(
+            "GET", f"{self.endpoints.core}/api/v1/devices/public-specs"
+        )
+        response.raise_for_status()
+        devices = response.json()["data"]
```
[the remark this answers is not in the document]
MohamedAminMallek: Good point! 
I think we can introduce the `_unauthenticated_request` once we have more than one unauthenticated requests

pasqal_cloud/client.py
MatthieuMoreau0: Did they patch python 3.8 to make this work? Because in the past this would raise an error

PR #140, opened 2024-10-01
[FEAT] Add deprecation warnings

Augustinio:
### Description

- Add deprecation warnings to __init__ of SDK
- Added `depreaction_date` to `_version.py` It will need to be updated for each version.

### Checklist

#### Documentation

#### Tests

- [x] Tests were run locally.


pasqal_cloud/__init__.py
```diff
@@ -54,6 +55,38 @@
...
+DEPRECATION_WARNING_PERIOD = timedelta(days=30)
+
+# Ensure DeprecationWarnings are not filtered out at runtime
+simplefilter("always", DeprecationWarning)
+
+
+def _check_sdk_version() -> None:
+    # Compare current date to deprecation date to check whether
+    # current version has fallen out of maintenance window.
+    current_date = datetime.now()
```
Mildophin, line 69: I am wondering about the UTC, as the releases were done in UTC+1 how does it work if someone from the US for example, call this method?

Augustinio: Added tzinfo

CHANGELOG.md
```diff
@@ -2,11 +2,17 @@
 
 All notable changes to this project will be documented in this file.
 
-## [0.12.3]
-_released `2024-10-02`_
+## [0.12.4] - 2024-10-07
+
+### Added
+
+- Deprecation warnings on init of SDK
```
Mildophin, line 9: Might be worth explaining it a bit more clearly to the user, as a developer point of view it seems very logical but maybe something like this could be clearer for non-maintainers:

Deprecation warnings added, now each version of pasqal-cloud is supported for 1 year after release. If your version is not supported anymore you will get a deprecation warning each time you use the SDK.

Mildophin (approved):

pasqal_cloud/__init__.py
```diff
@@ -54,6 +55,40 @@
 from pasqal_cloud.utils.responses import JobCancellationResponse, PaginatedResponse
 from pasqal_cloud.workload import Workload
 
+from ._version import __version__, deprecation_date
+
+DEPRECATION_WARNING_PERIOD = timedelta(days=30)
+
+# Ensure DeprecationWarnings are not filtered out at runtime
+simplefilter("always", DeprecationWarning)
```
MatthieuMoreau0, line 63: I'm not familiar with this, what does this do and what is the idea behind it?

MatthieuMoreau0 (approved):

pasqal_cloud/__init__.py
Augustinio: This is to ensure python doesn't filter out Deprecation warnings at runtime. I was encountering them fine in the tests but at runtime, locally the warnings didn't show up. This fixed it.

PR #147, opened 2024-11-08
[FEAT] Incorporate pulser-pasqal package in the pasqal-cloud repo

HGSilveri:
### Description
- Isolate `pasqal_cloud` package inside the `pasqal-cloud` directory
- Add the `pulser-pasqal` directory with the `pulser_pasqal` package

#### Remaining Tasks

- [x] Include `pulser-pasqal` in the `mypy` checks
- [x] Share a centralized version between the two packages
- [x] Update the "Publish" workflow to release both packages simultaneously
- [x] Setup TestPyPI and PyPI to use trusted publishing
- [x] Update the repo README

### Checklist

#### Documentation

#### Tests

- [x] Tests were run locally.


HGSilveri:
**Note for reviewers**: Reviewing one commit at a time is recommended on this one

oliver-gordon:
No, these don't reflect a direct change in the cloud release so changelog is OK without an update.

But this leaves me to assume you think we should bump the cloud version to twin with the pulser-pasqal lib in another MR?

if I've missed it please point me to it.

I think this looks OK

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.20.2]
+
+_release `2024-11-25`_
+
+- Synchronizes the version with `pulser-pasqal`. From now, both pacakges are released in tandem.
```
Mildophin, line 9: typo: packages*

pasqal-cloud/README.md
```diff
@@ -0,0 +1,341 @@
...
+| 0.7.0   | 2024-03-05   | 2025-03-05     |
+| 0.8.0   | 2024-04-15   | 2025-04-15     |
+| 0.9.0   | 2024-05-15   | 2025-05-15     |
+| 0.10.0  | 2024-06-05   | 2025-06-05     |
+| 0.11.0  | 2024-06-27   | 2025-06-27     |
+| 0.12.0  | 2024-09-03   | 2025-09-03     |
+| 0.12.1  | 2024-09-11   | 2025-09-11     |
+| 0.12.2  | 2024-09-11   | 2025-09-11     |
+| 0.12.3  | 2024-10-02   | 2025-10-02     |
+| 0.12.4  | 2024-10-09   | 2025-10-09     |
```
Mildophin, line 20: I would mean adding it here I believe

pyproject.toml
```diff
@@ -4,6 +4,8 @@ build-backend = "setuptools.build_meta"
 
 [tool.ruff]
 
+src = ["pasqal-cloud", "pulser-pasqal"]
```
Mildophin, line 7: I am not here it is needed as it i scanning the main directory, didn't it work properly ?

Mildophin:
> @oliver-gordon @MatthieuMoreau0 The version is an issue I didn't make a decision on. As it stands,
> 
>     * The two packages are set to share the same version number and the publishing workflow is triggered by pushes to main, so this will _attempt_ to publish both `pasqal-cloud` and `pulser-pasqal` to 0.20.1 simultaneously.
> 
>     * However, `pulser-pasqal` 0.20.1 already exists, so this will fail to some extent. Whether it would still succeed at publishing `pasqal-cloud` I don't know but I don't think we should test it.
> 
>     * So, I think we should update the version. My questions to you are then:
>       
>       * To what version?
>       * Is it fine to have a release on the CHANGELOG without changes? If not, what should I fill in there?
> 
> 
> EDIT: I guess we can also set the CI workflow to `skip-existing: true` for PyPI so that we publish only `pasqal-cloud` at 0.20.1

I am wondering if the bump to 0.21.0 would be clearer for the previous user of pulser pasqal ? It is a bit a breaking change in some way. I am not sure though.

It is fine yes if there is no feature added (or one changed). I guess it would be great to have the documentation about how to use pulser-pasqal feature somewhere in it.

pyproject.toml
HGSilveri: Without this the imports in the `tests` see `pasqal-cloud` and `pulser-pasqal` as third-party libraries and enforce an import order that, to me, doesn't make much sense.

HGSilveri:
> I am wondering if the bump to 0.21.0 would be clearer for the previous user of pulser pasqal ? It is a bit a breaking change in some way. I am not sure though.

For the user, there is no breaking change since the library itself did not change. I think 0.20.2 makes sense.

pyproject.toml
Mildophin: I see, it makes sense indeed

Mildophin:
I dont have much more things to bring to the table. Only two small points:

- As benjamin is working on the documentation, it has be be rebased on what he has done et maybe integrate it to github pages
- Date and version need to be updated

README.md
```diff
@@ -31,24 +32,30 @@ use pip:
 pip install pasqal-cloud
 ```
 
-If you wish to **install the development version of the pasqal-cloud from source** instead, do the following from within
+Likewise, with Python 3.9.0 or higher, `pulser-pasqal` is installed with:
```
AurelienNober, line 35: it's 3.8 in `setup.py` but I guess that's because `pulser-core` supports only greater than 3.9 since v1

AurelienNober (commented):
lgtm

let's see if we need to incorporate latest changes and update `pulser-core` to `~=1.3` on `pulser-pasqal`

.github/workflows/publish-to-testpypi-and-pypi.yml
AurelienNober, line 1: not related to this PR but I think triggering publish should be done on git tag rather than main branch to ease management of multiple ongoing major, minor, patch.

not a big deal, one can still code on a specific branch and publish manually anyway

HGSilveri:
> let's see if we need to incorporate latest changes and update `pulser-core` to `~=1.3` on `pulser-pasqal`

I thought about this but I figured that it would be good to have a first release that's compatible with older versions of `pulser-core`. We can then incorporate the changes in 1.3 and upgrade the dependency in the next release

AurelienNober:
> > let's see if we need to incorporate latest changes and update `pulser-core` to `~=1.3` on `pulser-pasqal`
> 
> I thought about this but I figured that it would be good to have a first release that's compatible with older versions of `pulser-core`. We can then incorporate the changes in 1.3 and upgrade the dependency in the next release

sounds good!

README.md
```diff
@@ -21,6 +21,7 @@ SDK to be used to access Pasqal Cloud Services.
 | 0.12.5  | 2024-11-12   | 2025-11-12     |
 | 0.12.6  | 2024-12-12   | 2025-12-12     |
 | 0.12.7  | 2025-01-14   | 2026-01-14     |
+| 0.20.2  | 2024-02-21   | 2026-02-21     |
```
MatthieuMoreau0, line 24: This timeline needs to be edited for pulser pasqal old versions that we still want to support

bbecquet:
Approving after checking that documentation still works fine with this restructuration :white_check_mark:

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.20.2]
+
+_release `2024-11-26`_
+
+- Synchronizes the version with `pulser-pasqal`. From now on, both packages are released in tandem.
+
```
Mildophin, line 10: this part is wrong: dates, and it's about to be merged, isn't merged yet

pulser-pasqal/README.md
```diff
@@ -0,0 +1,21 @@
...
+pip install pulser
+```
+
+will automatically install `pulser-pasqal`. If you wish to install it on its own, you can also run
+
+```bash
+pip install pulser-pasqal
+```
+
+Note that `pulser-core` is a requirement of `pulser-pasqal`, so it will be installed if it hasn't been already.
```
awennersteen, line 21: will this still be the behaviour after moving pulser-pasqal to this repo?

PR #161, opened 2025-01-28
[FEAT] Add MkDocs-based API documentation

bbecquet:
### Description

- Setup a documentation site, taking inspiration from the Qadence doc: same tool (MkDocs) + styling and config
- Adds some autogen'd API doc  based on docstrings, for SDK, batches, jobs, workloads and errors
- Fix many docstring syntax problems resulting in warnings or incorrectly parsed annotations 

Capture d’écran 2025-01-28 à 13 58 21

Depends on the following dependencies:
- `mkdocs`
- `mkdocstrings` && `mkdocstrings-python`
- `mkdocs-material`

Can be installed with `pip install -e .[docs]`

Then to launch the web server:

```shell
mkdocs serve
```

#### Remaining Tasks

- [x] ~~Properly declare dependencies~~
- [ ] Later: Fill some non-API content. Probably by moving some content from the Readme file
- [ ] Later: publish to GH-pages with GitHub action 

### Checklist

- [X] The title of the PR follows the right format: [{Label}] {Short Message}. Label examples: IMPROVEMENT, FIX, REFACTORING... Short message is about what your PR changes.

#### Tests

- [x] Tests were run locally.


awennersteen:
if you want to have the dependencies in the main `setup.py` you can add to `extras_require` like this:
```
extras_require={
        "docs": {
            "mkdocs==x.y.z",
            "mkdocs-material"
        }
```

Alternatively it's common to put a `requirements_docs.txt`, often inside the `docs` directory, with the requirements there.

MohamedAminMallek:
Very cool tool 👍 

For dependencies, I believe we can include a `docs` section in the `extras_require`.

Also, I noticed that classes without docstrings (for example, `ExceptionWithResponseContext`) are excluded from the documentation, which I think is fine.

MohamedAminMallek (approved):

bbecquet:
@awennersteen @MohamedAminMallek thanks for the review and the help about dependencies. I've listed them manually in `extra_requires` in `setup.py`, it works great :+1:

Probably they don't need such precise versions, but it's the way it's done for the dev ones so :shrug:

MatthieuMoreau0 (approved):
Thanks a lot for this!

MatthieuMoreau0:
It looks like the docstring of the `__init__` function is printed on the documentation. However for the batch, job and workload class this currently holds some information meant for the developpers of the library and not its users (see screenshot).

Is it possible to omit parts of a docstring using a flag or something?

image

bbecquet:
> It looks like the docstring of the `__init__` function is printed on the documentation. However for the batch, job and workload class this currently holds some information meant for the developpers of the library and not its users (see screenshot).
> 
> Is it possible to omit parts of a docstring using a flag or something?

@MatthieuMoreau0 I've pushed a [new commit](https://github.com/pasqal-io/pasqal-cloud/pull/161/commits/39ae8c27785cec4b727521309ebe1fdda66fbd54) to set the `ignore_init_summary` option to `true`, which solves the problem… if the comment has only two lines (?), otherwise the third is not considered part of the summary and not ignored.

I think there is some docstring convention at play here, does that ring a bell?

MatthieuMoreau0:
@bbecquet This doesn't sound like the right feature for our needs, I suggest we just move these "developers" comments out of the docstrings. That will do the trick

MatthieuMoreau0 (approved):

PR #163, opened 2025-02-04
[CI] Add GitHub action for auto-publishing docs

bbecquet:
### Description

This adds a GitHub Action to automatically update the MkDocs-based documentation site when the repo is updated:

This uses [this recipe](https://squidfunk.github.io/mkdocs-material/publishing-your-site/#with-github-actions) with the following changes:
  - restrict the trigger condition to pushes on the `main` branch (we don't use `master`) 
  - replace `pip install mkdocs-material` by `pip install .[docs]` to install all doc dependencies listed in `setup.py`
  - change the action name and the cache key prefix for something more explicit 

It uses of the [built-in `mkdocs gh-pages` command](https://www.mkdocs.org/user-guide/deploying-your-docs/#project-pages) that builds and push by default to `gh-pages` from another branch with doc sources.

Tested locally with [act](https://nektosact.com/), everything worked fine.

### Checklist

#### Documentation

### DIFF
--- a/.github/workflows/publish-doc-to-gh-pages.yml
+++ b/.github/workflows/publish-doc-to-gh-pages.yml
@@ -1,0 +1,28 @@
+name: Publish MkDocs documentation to gh-pages
+on:
+  push:
+    branches:
+      - main
+permissions:
+  contents: write
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Configure Git Credentials
+        run: |
+          git config user.name github-actions[bot]
+          git config user.email 41898282+github-actions[bot]@users.noreply.github.com
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.x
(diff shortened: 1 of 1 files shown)


Mildophin (approved):

MatthieuMoreau0 (approved):
LGTM!

Out of scope for this MR: so I notice we publish when merging to main. This makes sense since, afaict, only the latest version is documented. How much effort would it take to document each versions so that users could refer to the documentation of their version?

bbecquet:
@Mildophin @MatthieuMoreau0 thanks for the approval!

> Out of scope for this MR: so I notice we publish when merging to main. This makes sense since, afaict, only the latest version is documented. How much effort would it take to document each versions so that users could refer to the documentation of their version?

Some other Pasqal projects documented on MkDocs are using [mike](https://github.com/jimporter/mike), which does exactly this. The principle is to keep old doc versions built in the `gh-pages` branch, and only update the latest. Looks like a nice tool, I can try to set it up in a next PR :)

PR #164, opened 2025-02-04
[FIX] Fix access to sequence builder attribute for batch instances

MatthieuMoreau0:
### Description

An internal user has reported an issue when trying to access the sequence builder attribute from a batch instance. This MR fixes it.

See Slack thread: https://pasqalworkspace.slack.com/archives/C03A5T283C4/p1738691199579379?thread_ts=1738689449.189079&cid=C03A5T283C4

#### Breaking changes

No breaking change

### Checklist

#### Documentation

#### Tests

- [ ] Tests were run locally.


MatthieuMoreau0:
> Can we add a little test verifying the sequence builder field is accessible ? GTM otherwise

Yeah definitely on my TODO for this MR. Some tests need to be fixed as well, since this creates extra call to core

MohamedAminMallek (approved):

Mildophin:
@MatthieuMoreau0 not sure if we can add test verifying the problem that is fixed with this PR, from my understanding it looks more an E2E test. I fixed the problematic tests but I don't see how to test more in detail the property behaviour.

Mildophin:
@MatthieuMoreau0  indeed it was possible, the mock response of GET all batches endpoint had to be modified.

before my MR:

Screenshot 2025-03-20 at 10 30 12

And now test should be green

tests/test_batch.py
```diff
@@ -931,3 +934,14 @@ def test_cancel_batches_success(
...
+    def test_sequence_builder_field_is_accessible(
+        self,
+        mock_request: Any,
+    ):
+        """Simple test verifying the property sequence_builder
+        of a batch is accessible"""
+        response = self.sdk.get_batches()
+        first_batch = response.results[0]
+        assert first_batch.sequence_builder == self.pulser_sequence
+        assert mock_request.last_request.method == "GET"
```
MatthieuMoreau0, line 947: maybe assert that the get batch endpoint is called. I would amend the description to clarify that the get batches endpoint doesn't return the sequence builder, but the sequence builder can be accessed anyway and will trigger a call to the get batch endpoint

MatthieuMoreau0 (commented):
lgtm, can't approve since I contributed

MohamedAminMallek (approved):

PR #171, opened 2025-03-10
[FEAT] Create batch with device type

Augustinio:
### Description

- Added device type arg to batch creation parameters
- Raise warning when using deprecated `emulator` param

### Checklist

#### Documentation

#### Tests

- [ ] Tests were run locally.


pasqal-cloud/pasqal_cloud/__init__.py
```diff
@@ -169,13 +170,52 @@ def _get_batch(
...
+        if emulator is not None:
+            warn(
+                "Argument `emulator` is deprecated and will be removed in a"
+                " future version. Please use argument `device_type` instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            return DeviceTypeName(str(emulator))
+        if emulator is None and device_type is None:
+            return DeviceTypeName.FRESNEL
```
MatthieuMoreau0, line 189: The backend already takes care of the backwards compatibility for this, where it will keep extracting the device type from the sequence if both `emulator` and `device_type` are unset. I think here we can keep it as None but raise a deprecation warning saying that device_type should be defined explicitly

pasqal-cloud/pasqal_cloud/__init__.py
```diff
@@ -169,13 +170,52 @@ def _get_batch(
         except HTTPError as e:
             raise BatchFetchingError(e) from e
 
+    def _validate_device_type(
+        self,
+        emulator: Optional[EmulatorType],
```
MatthieuMoreau0, line 175: Can you look into raising a warning whenever a user instantiates a `EmulatorType` value and redirect the user to DeviceType instead.

Mark the enum as deprecated as well please.

MatthieuMoreau0 (approved):
A few comments about backwards compatibility but overall looks good

MatthieuMoreau0:
Make sure to update the documentation portal to reflect these changes as well

pasqal-cloud/pasqal_cloud/__init__.py
MatthieuMoreau0: mh ok I see

pasqal-cloud/pasqal_cloud/__init__.py
Augustinio: So I looked into deprecating the use of the Enum itself and it ended up quite convoluted with the update of multiple functions and it seemed really janky so I'd rather not implement it. I think a deprecation in the create batch is probably enough since it's the main point of use of the emulator. I've marked the class as deprecated though

pasqal-cloud/pasqal_cloud/__init__.py
```diff
@@ -169,13 +170,52 @@ def _get_batch(
...
+        if emulator is None and device_type is None:
+            return DeviceTypeName.FRESNEL
+        return device_type
+
+    def _validate_open(
+        self,
+        complete: Optional[bool],
+        open: Optional[bool],
+    ) -> bool:
+        if complete is not None and open is not None:
```
oliver-gordon, line 197: nit: can be `if x and y` ?

pasqal-cloud/pasqal_cloud/__init__.py
```diff
@@ -169,13 +170,52 @@ def _get_batch(
...
+            raise OnlyCompleteOrOpenCanBeSet
+        if complete is not None:
```
oliver-gordon, line 199: nit: `if complete`?

pasqal-cloud/pasqal_cloud/device/emulator_types.py
```diff
@@ -2,6 +2,11 @@
 
 
 class EmulatorType(StrEnum):
+    """
+    Enum for the emulator type, deprecated.
```
oliver-gordon, line 6: `is deprecated`?

pasqal-cloud/pasqal_cloud/errors.py
```diff
@@ -215,3 +215,15 @@ class OnlyCompleteOrOpenCanBeSet(BaseException):
 
     def __init__(self) -> None:
         super().__init__("Only complete or open can be set when creating a batch.")
+
+
+class OnlyEmulatorOrDeviceTypeCanBeSet(BaseException):
```
oliver-gordon, line 220: nit: maybe reduce the name here `InvalidDeviceTypeSet`?

Means we can change the error message later without needing to change the class name and we can leave it a tad more abstract in case something changes

pasqal-cloud/pasqal_cloud/__init__.py
Augustinio: no because we want to differentiate between None and False

pasqal-cloud/pasqal_cloud/__init__.py
Augustinio: same thing

Mildophin (approved):

PR #172, opened 2025-03-31
[FEAT] Add retries on transient errors on authenticated requests

Mildophin:
### Description

I introduce 5 exponential backoff retries on all authenticated requests.

I added tests verifying the most important get endpoints but not all of them (retry, close batch, get all jobs or get all batches are not tested) but they should be more or less because al the get endpoints using authenticated requests are now handled by a retry on Timeout and Connection Errors + some HTTP code (408, 425, 429, 500, 502, 504).

### Checklist

#### Documentation

#### Tests

- [ ] Tests were run locally.


tests/test_batch.py
```diff
@@ -959,3 +961,150 @@ def test_sequence_builder_field_is_accessible(
...
+        mock_request,
+        exception,
+        expected_exception,
+    ):
+        """Test that GET batch call retries on transient errors
+        after we create a batch."""
+        mock_request.reset_mock()
+
+        # Register the URI with the appropriate exception
+        exc = exception() if callable(exception) else exception
```
MatthieuMoreau0, line 990: I don't understand why some cases need a callable and some cases don't since you call it here right away

tests/test_batch.py
```diff
@@ -959,3 +961,150 @@ def test_sequence_builder_field_is_accessible(
...
+        assert (
+            mock_request.last_request.path
+            == f"/core-fast/api/v2/batches/{self.batch_id}"
+        )
+        assert mock_request.last_request.matcher.call_count == 6
+
+    @pytest.mark.parametrize(
+        ("exception", "expected_exception"),
+        [
+            (requests.Timeout, requests.Timeout),
```
MatthieuMoreau0, line 1019: So this is not directly related to your MR but I think it would be better if we raised BatchFetchingError for all types of network related errors as well. THis would make it easier for users to understand which step went wrong

MatthieuMoreau0 (approved):
Thanks for adding a bunch of test. Maybe we could add an extra test to verify we dont retry POST requests (e.g. batch creation) in case of 500 or connection error

tests/test_batch.py
Mildophin: the 500 error needs to be called.

Another option is that I modify the test to get the same architecture as test_sdk_retry_on_exception, I chose this one because it allowed to parametrize only on one field (exc) instead of one for status code and one for exceptions.

pasqal-cloud/pasqal_cloud/client.py
```diff
@@ -140,19 +145,37 @@ def _authenticated_request(
...
-        resp.raise_for_status()
+        headers = {
+            "content-type": "application/json",
+            "User-Agent": self.user_agent,
+        }
+
+        if method == "GET":
+            request_with_retry = retry_http_error(
+                max_retries=5,
+                retry_status_code={408, 425, 429, 500, 502, 504},
```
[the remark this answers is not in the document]
Mildophin: Exactly

tests/test_batch.py
MatthieuMoreau0: I don't get why the pytest parameter cannot simply be `requests.HTTPError("500 Server Error",response=mock_500_http_error_response())`

Mildophin: Indeed it seems to work, i might have over-engineered it a bit too much, thx for spotting it

pasqal-cloud/pasqal_cloud/client.py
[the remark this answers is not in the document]
MohamedAminMallek: > Here is why Post or patch aren't changed

I may be missing something, but it seems that you have removed the retry mechanism for requests other than GET. fyi, the retry was originally introduced mainly for POST requests when Core was experiencing issues scheduling EMU jobs through DGX. However, this may no longer be necessary.

That said, I’m wondering—why didn’t you use the retry_http_error decorator with retry_exceptions directly on the get_job/batch method instead of introducing this if-else logic in the _authenticated_request method?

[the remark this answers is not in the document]
MohamedAminMallek: Also wondering if you evaluated the risk of adding retry_exceptions to all requests, since it would be the simplest solution?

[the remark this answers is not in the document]
Mildophin: I put back the retries on all the authenticated requests as discussed.

MohamedAminMallek (approved):

pasqal-cloud/pasqal_cloud/client.py
```diff
@@ -140,19 +145,42 @@ def _authenticated_request(
...
+
+            resp = request_with_retry(
+                method,
+                url,
+                json=payload,
+                timeout=TIMEOUT,
+                headers=headers,
+                auth=self.authenticator,
+                params=params,
+            )
```
MohamedAminMallek, line 183:
```suggestion
        if method == "GET":
            request_with_retry = retry_http_error(
                max_retries=5,
                retry_status_code={408, 425, 429, 500, 502, 504},
                retry_exceptions=(requests.ConnectionError, requests.Timeout),  # type: ignore
            )(self._request_with_status_check)
        else:
            request_with_retry = retry_http_error(
                max_retries=5,
                retry_status_code={408, 425, 429, 500, 502, 504},
            )(self._request_with_status_check)

        resp = request_with_retry(
            method,
            url,
            json=payload,
            timeout=TIMEOUT,
            headers=headers,
            auth=self.authenticator,
            params=params,
        )
```

Mildophin: ah yes mb

PR #175, opened 2025-04-29
[FEAT] Introduce tags for batches

Mildophin:
### Description

With this PR a batch can be created with tags and an existing batch can a be added tags, either via the SDK client or by manipulating the Batch instance itself.

#### Remaining Tasks

Documentation will be added in the next PR where tags will be added to batch filters.

### Checklist

#### Documentation

#### Tests

- [ ] Tests were run locally.

### DIFF
--- a/pasqal-cloud/pasqal_cloud/__init__.py
+++ b/pasqal-cloud/pasqal_cloud/__init__.py
@@ -38,6 +38,7 @@ from pasqal_cloud.errors import (
     BatchClosingError,
     BatchCreationError,
     BatchFetchingError,
+    BatchTagsSettingError,
     DeviceSpecsFetchingError,
     InvalidDeviceTypeSet,
     JobCancellingError,
@@ -219,6 +220,7 @@ class SDK:
         configuration: Optional[BaseConfig] = None,
         wait: bool = False,
         fetch_results: bool = False,
+        tags: Optional[list[str]] = None,
     ) -> Batch:
         """Create a new batch and send it to the API.
 
@@ -240,7 +242,7 @@ class SDK:
                 terminated
             fetch_results (deprecated): Whether to wait for the batch to
                 be done and fetch results
-
+            tags: List of keywords that can be used to refine the Batch search
 
         Returns:
             Batch: The new batch that has been created in the database.
@@ -267,6 +269,7 @@ class SDK:
             "jobs": jobs,
             "open": open,
             "device_type": device_type,
+            "tags": tags,
         }
         # The configuration field is only added in the case
         # it's requested
@@ -773,3 +776,27 @@ class SDK:
             return self._client.get_device_specs_dict()
         except HTTPError as e:
             raise DeviceSpecsFetchingError(e) from e
+
+    def set_tags(
+        self, batch_id: str, tags_to_add: list[str], tags_to_remove: list[str]
+    ) -> Batch:
+        """Set tags to an existing batch by adding or removing them.
+
+        Args:
+            batch_id: Batch id.
+            tags_to_add: the tags to add to the batch.
+            tags_to_remove: the tags to remove from the batch.
+
+        Returns:
+            batch: The updated batch with tags.
+
+        Raises:
+            BatchTagsSettingError: If setting tags to a batch failed.
+        """
+
+        try:
+            resp = self._client.set_batch_tags(batch_id, tags_to_add, tags_to_remove)
+        except HTTPError as e:
+            raise BatchTagsSettingError(e)
+
+        return Batch(**resp, _client=self._client)
--- a/pasqal-cloud/pasqal_cloud/__init__.py
+++ b/pasqal-cloud/pasqal_cloud/__init__.py
@@ -38,7 +38,7 @@ from pasqal_cloud.errors import (
     BatchClosingError,
     BatchCreationError,
     BatchFetchingError,
-    BatchTagsSettingError,
+    BatchSetTagsError,
     DeviceSpecsFetchingError,
     InvalidDeviceTypeSet,
     JobCancellingError,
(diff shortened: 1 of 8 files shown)


pasqal-cloud/pasqal_cloud/client.py
```diff
@@ -397,3 +397,16 @@ def get_public_device_specs(self) -> Dict[str, str]:
         response.raise_for_status()
         devices = response.json()["data"]
         return {device["device_type"]: device["specs"] for device in devices}
+
+    def set_batch_tags(
```
oliver-gordon, line 401: Split this up, no one function in the SDK needs to do two things. `Add_tags` and `Set_tags`.

bbecquet: I don't know. "Setting" is not two things at the same time, it's just replacing the full list at once. It makes the API and implementation simpler and naturally idempotent. And it's actually already implemented like that https://docs.pasqal.com/cloud/api/core/operations/set_batch_tags_api_v1_batches__batch_id__tags_patch/ 

But I don't remember what we said about that @MatthieuMoreau0? Is the current API not compliant with your design principles?

oliver-gordon: While it's performing one request, it feels like a user is doing two things. I know how it's implemented but the SDK doesn't need to mirror this.

My thought behind this is it's simpler and more intuitive to have `.add()` or `.remove()` to perform add/remove actions than it is `update(remove=..., add=...)`.

bbecquet: @oliver-gordon sorry, my bad, I didn't read the API I linked carefully, didn't notice it was a single `set` endpoint but with a `add`/`remove` payload. Indeed it's the worst of both ways, too complex both for server and client IMO.

I thought it was just a `set` action, taking a single array as payload, and just overwriting the current one with it, not bothering at all about `add` / `remove`.

What would you think about such API?

MatthieuMoreau0: As a a general note, the tags API has never been used so we can make changes to its behaviour now if we don't like how it works. 

In the current API is that there is an "add" and "remove" field in the payload, tags are added and removed accordingly. imo this is unpractical and we could instead simply have a "set_tag" logic where the list is overriden by the list passed as argument; this sounds both simpler to implement for us and to use for the end user.

If you guys wanna keep the current API then I like Oliver's suggestion of separating the two actions in the SDK

oliver-gordon: The API is practical because it was designed around the idea that the front end can make a singular request to modify all the tags from a UI modal at once instead of having to make multiple requests. There was no thought put in for the SDK to need this what so ever. Web/API/SDK interfaces all have different criteria.

If we want to split the tags API now so we have `/add` and `/remove`, then I'd jump to it.

The rest of our SDK does not have interfaces like this one, we should be more consistent. 

Or maybe we can release a new version and begin merging other functions 
```python
def create_batch_or_resubmit_batch(create: bool=True, batch_to_resubmit: UUID | None = None, another_unnecessary_parameter:dict[str, Iterable[int | bool | float | uint32 | object Object] | None = None) 
```


The cheaper win here is to just have two functions in the SDK

bbecquet: > The API is practical because it was designed around the idea that the front end can make a singular request to modify all the tags from a UI modal at once instead of having to make multiple requests

I get the concern, but not sure the API should be designed with a precise front UI in mind, and actually the chosen API is more complex even for this UI (and for a backend) than having an overwrite-all `set` endpoint :D

In any case, it's not a big deal nor is this endpoint critical, so we can do both, and if this is less work to keep the current endpoint, no blocker for me. I just like consistency and overall API desgin discussions :)

bbecquet: Can't approve but thanks @Mildophin for the change, seems easier for everybody this way <3

pasqal-cloud/pasqal_cloud/__init__.py
```diff
@@ -219,6 +220,7 @@ def create_batch(
         configuration: Optional[BaseConfig] = None,
         wait: bool = False,
         fetch_results: bool = False,
+        tags: Optional[list[str]] = None,
```
MohamedAminMallek, line 223:
```suggestion
        tags: list[str] | None = None,
```

pasqal-cloud/pasqal_cloud/__init__.py
```diff
@@ -773,3 +778,28 @@ def get_device_specs_dict(self) -> Dict[str, str]:
...
+
+        Args:
+            batch_id: Batch id.
+            tags: the tags defining the batch
+
+        Returns:
+            batch: The updated batch with newly set tags.
+
+        Raises:
+            BatchTagsSettingError: If setting tags to a batch failed.
```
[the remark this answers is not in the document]
MohamedAminMallek:
```suggestion
            BatchSetTagsError: If setting tags to a batch failed.
```

pasqal-cloud/pasqal_cloud/batch.py
```diff
@@ -76,6 +78,7 @@ class Batch(BaseModel):
     parent_id: Optional[str] = None
     configuration: Union[BaseConfig, Dict[str, Any], None] = None
     _sequence_builder: Optional[str] = None
+    tags: Optional[list[str]] = None
```
MohamedAminMallek, line 81:
```suggestion
    tags: list[str] | None = None
```

MohamedAminMallek (approved):

pasqal-cloud/pasqal_cloud/__init__.py
MohamedAminMallek: Ah! Since we still support Python 3.9, this might not work

pasqal-cloud/pasqal_cloud/__init__.py
```diff
@@ -273,6 +275,9 @@ def create_batch(
         if configuration:
             req.update({"configuration": configuration.to_dict()})  # type: ignore[dict-item]
 
+        if tags:
+            req.update({"tags": tags})  # type: ignore[dict-item]
```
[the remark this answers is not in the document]
Mildophin: Without it I am getting a mypy error:

`Expected type 'str | None | list[CreateJob] | bool | DeviceTypeName' (matched generic type '_VT'), got 'list[str]' instead `


but not sure where it comes from, as it is only sent through the authentificated request and it is like this:

payload: Optional[Union[Mapping, Sequence[Mapping], Sequence[str]]]


if anyone has an idea

PR #176, opened 2025-05-02
[FEAT] Add tags to batch filters

Mildophin:
### Description

Simple MR adding tags to batch filters, allowing users to search batches by a specific tag.

#### Related PRs in other projects (PASQAL developers only)

https://github.com/pasqal-io/pasqal-cloud/pull/175

### Checklist

#### Documentation

#### Tests

- [ ] Tests were run locally.

### DIFF
--- a/docs/usage/batches_jobs.md
+++ b/docs/usage/batches_jobs.md
@@ -2,8 +2,8 @@ A batch is a group of jobs with the same sequence that will run on the same QPU.
 
 ## Create a batch of jobs
 
-The package main component is a Python object called [`SDK`][pasqal_cloud.SDK] which can be used to create a [
-`Batch`][pasqal_cloud.Batch].
+The package main component is a Python object called [`SDK`][pasqal_cloud.SDK] which can be used to create a
+[`Batch`][pasqal_cloud.Batch].
 
 For each [`Job`][pasqal_cloud.Job] of a given batch, you must set a value for each variable, if any, defined in your
 sequence.
@@ -57,16 +57,6 @@ batch.add_jobs([job2, job3], wait=True)
 batch.close()
 ```
 
-You can assign multiple tags to your batches when creating them to help organize and retrieve them later.
-
-```python
-batch = sdk.create_batch(
-    serialized_sequence,
-    [job1, job2],
-    tags=["special_experiment"]
-)
-```
-
 You can also choose to run your batch on an emulator using the argument `device_type_name`.
 For using a basic single-threaded QPU emulator that can go up to 10 qubits, you can specify the "EMU_FREE" emulator:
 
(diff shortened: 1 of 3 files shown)


docs/usage/batches_jobs.md
```diff
@@ -2,9 +2,11 @@ A batch is a group of jobs with the same sequence that will run on the same QPU.
 
 ## Create a batch of jobs
 
-The package main component is a Python object called [`SDK`][pasqal_cloud.SDK] which can be used to create a [`Batch`][pasqal_cloud.Batch].
+The package main component is a Python object called [`SDK`][pasqal_cloud.SDK] which can be used to create a [
+`Batch`][pasqal_cloud.Batch].
 
-For each [`Job`][pasqal_cloud.Job] of a given batch, you must set a value for each variable, if any, defined in your sequence.
+For each [`Job`][pasqal_cloud.Job] of a given batch, you must set a value for each variable, if any, defined in your
+sequence.
```
bbecquet, line 9: Maybe it's a manual mistake, however I suspect a formatting or IDE setup that cut lines above n chars is responsible for this change. In any case you should disable it on Markdown files, as apparently it can cut right in the middle of an expression (here a link), breaking the rendering:

image

docs/usage/batches_jobs.md
```diff
@@ -55,6 +57,16 @@ batch.add_jobs([job2, job3], wait=True)
 batch.close()
 ```
 
+You can assign multiple tags to your batches when creating them to help organize and retrieve them later.
+
+```python
+batch = sdk.create_batch(
+    serialized_sequence,
+    [job1, job2],
+    tags=["special_experiment"]
```
bbecquet, line 66: Suggestion: the comment above mentions multiple tags, so it would be nice to have two of them in the array of this example to illustrate it.

docs/usage/batches_jobs.md
Mildophin: It was automatic, I didn't get why my IDE did it though, I thought it was wrongly formatted initially

docs/usage/batches_jobs.md
Mildophin: sure, it makes sense, let me add it

docs/usage/batches_jobs.md
bbecquet: BTW, in case you don't know how to check that live, you have to run `mkdocs serve` in the project root :)

docs/usage/batches_jobs.md
```diff
@@ -55,6 +57,16 @@ batch.add_jobs([job2, job3], wait=True)
 batch.close()
 ```
 
+You can assign multiple tags to your batches when creating them to help organize and retrieve them later.
```
MatthieuMoreau0, line 60: nit: This is debatable, but I would suggest moving this example out of this section which is really focused around batch creation. Tags's purpose is to conveniently retrieve a list of batches so I would move this in the `Get a list of batches` section.

I would also put in the same snippet an example which creates a couple batches with different/shared tags and then a get_batches call to retrieve some of them. Maybe also in that same snippet show the usage of the `set_tags` method as well. what do you think?

Mildophin:
> This couldn't just be done in the same MR? It looks like you're merging it into the other open one anyway

I prefered to split, personal preference even it could have been done

docs/usage/batches_jobs.md
bbecquet: I confirm it's fixed with the [ on the right line, so resolving.

But I still think it would be good to avoid these reformattings adding breaks in the middle of a sentence, formatters shouldn't tinker with that.

MatthieuMoreau0 (approved):

docs/usage/batches_jobs.md
```diff
@@ -73,6 +75,34 @@ for job in batch.ordered_jobs:
...
 
+## Set tags to batches
+
+You can assign multiple tags to your batches when creating them to help organize and retrieve them later.
+
+```python
+batch = sdk.create_batch(
+    serialized_sequence,
+    [job1, job2],
+    tags=["special_experiment", "first_try"]
```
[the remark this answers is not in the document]
Mildophin: sure, happy to change, I did not have any idea of the names they could have

PR #177, opened 2025-05-15
[FEAT] Add support for backend_configurations

oliver-gordon:
### Description

Add argument to enable backend_configurations in the SDK. These are serialised config objects from Pulser/Corresponding emulation library.

An example will be released in our docs alongside this request.

### Checklist

#### Documentation

#### Tests

- [ ] Tests were run locally.

### DIFF
--- a/pasqal-cloud/pasqal_cloud/batch.py
+++ b/pasqal-cloud/pasqal_cloud/batch.py
@@ -43,7 +43,7 @@ class Batch(BaseModel):
         status: Status of the batch. Possible values are:
             PENDING, RUNNING, DONE, CANCELED, TIMED_OUT, ERROR, PAUSED.
         _client: A Client instance to connect to PCS.
-        _ordered_jobs: List of all the jobs added to the batch,
+        ordered_jobs: List of all the jobs added to the batch,
             ordered by creation time.
         jobs_count: Number of jobs added to the batch.
         jobs_count_per_status: Number of jobs per status.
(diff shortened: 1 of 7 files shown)


pasqal-cloud/pasqal_cloud/batch.py
```diff
@@ -42,18 +42,20 @@ class Batch(BaseModel):
...
             PENDING, RUNNING, DONE, CANCELED, TIMED_OUT, ERROR, PAUSED.
-        webhook: Webhook where the job results are automatically sent to.
         _client: A Client instance to connect to PCS.
-        sequence_builder: Pulser sequence of the batch.
-        start_datetime: Timestamp of the time the batch was sent to the QPU.
-        end_datetime: Timestamp of when the batch process was finished.
-        device_status: Status of the device where the batch is running.
-        jobs (deprecated): Dictionary of all the jobs added to the batch.
-        ordered_jobs: List of all the jobs added to the batch,
+        _ordered_jobs: List of all the jobs added to the batch,
```
Mildophin, line 46: It is not a private attribute

pasqal-cloud/pasqal_cloud/batch.py
```diff
@@ -42,18 +42,20 @@ class Batch(BaseModel):
...
             ordered by creation time.
         jobs_count: Number of jobs added to the batch.
         jobs_count_per_status: Number of jobs per status.
+        webhook: Webhook where the job results are automatically sent to.
+        start_datetime: Timestamp of the time the batch was sent to the QPU.
+        end_datetime: Timestamp of when the batch process was finished.
+        device_status: Status of the device where the batch is running.
+        parent_id: ID from previous batc if a batch was resubmitted for execution.
         configuration: Further configuration for certain emulators.
+        backend_configuration: serialised config object for emulation runtimes.
```
MohamedAminMallek, line 56: Having both `configuration` and `backend_configuration` is a bit confusing for users, I think. I assume this will be clarified in the documentation, but could we add more details here about the difference between them, when to use backend_configuration over configuration, and maybe mention that backend_configuration is a serialized string representing an object from Pulser?

Also, if I understand correctly, we are deprecating configuration in favor of backend_configuration, right? For now, this applies only to EMU_MPS, but the plan is to do the same for all emulators, correct?

MohamedAminMallek (approved):

pasqal-cloud/pasqal_cloud/batch.py
oliver-gordon: It seems I copied this from the actual attribute (which is)

```
_ordered_jobs: Optional[List[Job]] = PrivateAttr(default=None)
```

PR #189, opened 2025-06-03
New device_type: FRESNEL_RD

MohamedAminMallek:
Added support for a new device_type: `FRESNEL_RD`


MatthieuMoreau0 (approved):
Out of curiosity have you checked if there is some validation on the values used?

Mildophin (approved):

PR #196, opened 2025-07-03
[FEAT] Disable SSL verification with env variable

MatthieuMoreau0:
### Description

Introduce environment variable`PASQAL_SKIP_SSL_VERIFY` to skip ssl verification when making requests to our APIs. This is required by some clients who want to use the SDK behind a proxy for which they don't have access to the certificate when testing.

Set this environment variable to true (any non falsy value will do) when running your script if you want to skip SSL verifications.

We use the auth0 library to get a token and this library does not expose any way to skip SSL verification so I implemented a custom `InsecureAuth0TokenProvider` which bypasses SSL verification when fetching the token.

### Checklist

#### Documentation

#### Tests

- [x] Tests were run locally.


pasqal-cloud/pasqal_cloud/authentication.py
```diff
@@ -146,8 +151,33 @@ def _query_token(self) -> dict[str, Any]:
...
+                "realm": self.auth0.realm,
+                "client_secret": "",
+                "scope": AUTH0_TOKEN_PROVIDER_SCOPE,
+                "audience": self.auth0.audience,
+                "grant_type": AUTH0_TOKEN_PROVIDER_GRANT_TYPE,
+            },
+            verify=False,
+        )
+        response.raise_for_status()
+        return response.json()
```
AurelienNober, line 183: We could modify the `_query_token` method above to use this raw requests if the env skip ssl verify is defined. From user point of view, it will be simpler, you only have to set the env var once and everything is done for you.

Considering the short timeframe we have, I leave it up to you. It will be a beta anyway.

AurelienNober (approved):
One comment otherwise LGTM
Tested and reproduced in notebooks with containerized env client/proxy like you -> 🟢 
Probably we could use a `Session` to avoid the widespread `verify=not stuff` and simplify things.

Can you add an issue for the token generation in the API documentation?
Can you also add an issue to add the proxy and ssl typical problems to our doc? It's not directly linked to Pasqal software but it will help the users anyway which is what matters.

No need for another round of review from me tmw

pasqal-cloud/pasqal_cloud/authentication.py
MatthieuMoreau0: Yeah I considered this; before doing this we would need to refine a bit the errors. The error messages are currently more detailed with the auth0 client than with the plain request. I want to release this ASAP so they can test and we can refactor down the road.

PR #200, opened 2025-07-04
Merge main into dev after conflicts resolution

AurelienNober:


AurelienNober:
honestly I will push directly to dev

PR #209, opened 2025-09-04
[DOC] Create tutorial to use QPU from OVH platform

MatthieuMoreau0:
### Description

New tutorial to run an AFM state preparation on a QPU using the OVHConnection. I duplicated the myQLM tutorial and adapted it for OVH environment.

Note:
With the previous Fresnel specs, there is an issue when generating the sequence. 

```
ValueError: The pulse's detuning values go out of the range allowed for the chosen channel.
```

We updated the specs today which solves it so it's okay for now but it would be nice if we could make the pulse Sequence build in this sequence compatible with any Fresnel specs in the future


a-corni (approved):
LGTM !

PR #220, opened 2025-09-18
Fixes: Measuring basis of sequence built with job's variables + deserializing results

### LINKED ISSUE #218: PasqalCloud doesn't handle parametrized sequences
When submitting a parametrized to a QPU via PasqalCloud, one does:
```
import pulser
from pulser_pasqal import PasqalCloud
conn = PasqalCloud(...)
device = conn.fetch_available_devices()["my_device"]
param_seq = pulser.Sequence(device.calibrated_register_layouts["TriangularLatticeLayout(61, 5.0µm)"].define_register(21, 26, 35, 39, 34, 25), device)
param_seq.declare_channel("rydberg_global", "rydberg_global")
t = param_seq.declare_variable("t")
param_seq.add(pulser.Pulse.ConstantPulse(t, 1.0, 0.0, 0), "rydberg_global")
backend = pulser.QPUBackend(param_seq, conn)
remote_results = backend.run(job_params={"runs":100, "variables":{"t", 100}}, wait=False)
```
Once the batch is done, remote_results[0] returns a pulser.Result. However, this returns an error, because PasqalCloud._query_job_progress tries to fetch the measurement basis of the parametrized sequence `param_seq` (which raises an error), instead of the built sequence (`param_seq.build(t=100)` or `param_seq.build(**{'t':100})`).

Lines of the issue: https://github.com/pasqal-io/pasqal-cloud/blob/main/pulser-pasqal/pulser_pasqal/pasqal_cloud.py#L191-L194

A solution would be to be able to retrieve the variables for each job, build the sequence with these variables, and fetch the measurement basis afterwards (I don't see this option in the documentaiton however https://docs.pasqal.com/cloud/pasqal-cloud/api/batch/#pasqal_cloud.batch.Batch.jobs)

### LINKED ISSUE #221: EmuFreeBackendV2 does not return a pulser.Backend.Results
Users of EmuFreeBackendV2 (using QutipBackendV2) and EmuMPSBackend (using EmuMPS) expect a pulser.backend.Results from a `run`. pulser.backend.Results support serialization. This is currently being used in jobs submitted to the cloud via EmuMPSBackend: when the job is finished, job.full_result contains a `serialized_result` field.

This field is currently missing in jobs submitted to the cloud via EmuFreeBackendV2. Only "raw" and "counter" fields are defined in job.full_result. This is an issue if the users are asking for other observables to be measured than BitString, since the outcome will not be returned.

Solution:

- The MR #220 contains a fix that enables to return pulser.backend.Results if job.full_result contains a `serialized_result`: This currently enables users to get a pulser.backend.Results if submitting to EmuMPSBackend, but not to EmuFreeBackendV2.
- On the cloud side, associate with the key `serialized_result` in the data of a results of a job having run on QutipBackendV2, the result of the emulation in its serialized format (with to_abstract_repr).

a-corni:
### Description

When fetching results, the measurement basis is computed from a sequence built from the batch's Sequence with job's variables, instead of the batch's Sequence only, that can be parametrized, in which case `Sequence.get_measurement_basis` would return an error.

#### Remaining Tasks

Couldn't find a test that was reproducing error from the issue #218 

#### Additional merge criteria

It's a hotfix - up to you to decide

#### Tests

- [x] Tests were run locally.

Closes #218 
Partially closes #221

### DIFF
--- a/pulser-pasqal/pulser_pasqal/pasqal_cloud.py
+++ b/pulser-pasqal/pulser_pasqal/pasqal_cloud.py
@@ -203,14 +203,18 @@ class PasqalCloud(RemoteConnection):
             size: int | None = None
             if vars and "qubits" in vars:
                 size = len(vars["qubits"])
-            if "serialised_results" in job.full_result:
-                # preferably, has a serialized pulser Results
-                results[job.id] = (
-                    JobStatus[job.status],
-                    Results.from_abstract_repr(job.full_result["serialised_results"]),
-                )
-            elif job.result is not None:
+            if job.full_result is not None:
+                if "serialised_results" in job.full_result:
+                    # preferably, has a serialized pulser Results
+                    results[job.id] = (
+                        JobStatus[job.status],
+                        Results.from_abstract_repr(
+                            job.full_result["serialised_results"]
+                        ),
+                    )
+                    continue
                 # There is always a counter
+                assert job.result is not None
                 results[job.id] = (
                     JobStatus[job.status],
                     SampledResult(
@@ -219,8 +223,8 @@ class PasqalCloud(RemoteConnection):
                         bitstring_counts=job.result,
                     ),
                 )
-            else:
-                results[job.id] = (JobStatus[job.status], None)
+                continue
+            results[job.id] = (JobStatus[job.status], None)
         return results
 
     @backoff_decorator


tests/test_pulser_pasqal.py
```diff
@@ -410,6 +410,14 @@ def test_submit(fixt_pasqal_cloud, parametrized, emulator, mimic_qpu, seq, mock_
     remote_results = fixt_pasqal_cloud.pasqal_cloud.submit(
         seq, job_params=job_params, batch_id="open_batch"
     )
+    assert remote_results.get_available_results() == {
```
MatthieuMoreau0, line 413: I tried to reproduce the bug in the test by adding this, but somehow the test passes. Do you know why?

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
```diff
@@ -191,18 +192,25 @@ def _query_job_progress(
...
+            vars = job.variables or {}
+            meas_basis = (
+                seq_builder.build(**vars) if vars else seq_builder
+            ).get_measurement_basis()
             size: int | None = None
             if vars and "qubits" in vars:
                 size = len(vars["qubits"])
             if job.result is None:
                 results[job.id] = (JobStatus[job.status], None)
+                continue
```
MatthieuMoreau0, line 208: If the user doesn't define bitstrings observables, job.result will be None. In that case, don't we want instead to try to return the full_results as below ?`(JobStatus[job.status], Results.from_abstract_repr(job.full_result["serialised_results"]))`

sgrava:
Hi all,
it seems related, so posting it here.
I'm having these issues on pulser 1.6.1 and pulser-cloud 0.20.5

1. on both `EmuFreeBackendV2` and `EmuMPSBackand`, the returned results have 0 total duration and null UUID:
2. on `EmuMPSBackand`, there are no results. 

This is what I get from a simple sequence run asking for 100 bitstrings on `EmuMPSBackend`:
 ```
(SampledResult(atom_order=('0', '1'), total_duration=0, _results={UUID('00000000-0000-0000-0000-000000000000'): [{}]}, _times={UUID('00000000-0000-0000-0000-000000000000'): [1.0]}, _aggregation_methods={UUID('00000000-0000-0000-0000-000000000000'): <AggregationMethod.BAG_UNION: 3>}, _tagmap={'bitstrings': UUID('00000000-0000-0000-0000-000000000000')}, meas_basis='ground-rydberg', bitstring_counts={}, evaluation_time=1.0),)
```

Are 1 and 2 both solved here?

a-corni:
@MatthieuMoreau0 @MohamedAminMallek I let you two have a look

MohamedAminMallek (approved):

MatthieuMoreau0 (approved):
LGTM, thank you!

@MohamedAminMallek Can we schedule a release shortly for this bugfix? Let's make sure to update the changelog when doing that

PR #263, opened 2026-03-26
[CI/CD]: trigger publish workflows on tag push instead of push to main

Fenkiou:
### Description

Change release trigger from push on main branch to tags push instead.

### Checklist

#### Documentation

#### Tests

- [x] Tests were run locally.

### DIFF
--- a/.github/workflows/publish-doc-to-gh-pages.yml
+++ b/.github/workflows/publish-doc-to-gh-pages.yml
@@ -10,6 +10,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - name: Verify tag matches VERSION.txt
+        run: |
+          # strips refs/tags/ prefix from GITHUB_REF
+          tag="${GITHUB_REF#refs/tags/}"
+          version=$(cat VERSION.txt)
+          if [ "$tag" != "$version" ]; then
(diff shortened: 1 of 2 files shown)


MohamedAminMallek (commented):
Tags will trigger the release, but we will still rely on VERSION.txt. As discussed before, ideally, we would stop using VERSION.txt and dynamically retrieve the version from the tag, but this may require some additional work.

To avoid inconsistencies between the pushed tag and the content of VERSION.txt, let’s add a validation step before publishing a new version. This step should verify that the pushed tag matches the content of VERSION.txt, otherwise, the release should fail.

@Fenkiou, what do you think?

Fenkiou:
> Tags will trigger the release, but we will still rely on VERSION.txt. As discussed before, ideally, we would stop using VERSION.txt and dynamically retrieve the version from the tag, but this may require some additional work.
> 
> To avoid inconsistencies between the pushed tag and the content of VERSION.txt, let’s add a validation step before publishing a new version. This step should verify that the pushed tag matches the content of VERSION.txt, otherwise, the release should fail.
> 
> @Fenkiou, what do you think?

Very good point I'll do that 👍

Fenkiou:
@MohamedAminMallek done :), I've tested it here: https://github.com/Fenkiou/pasqal-cloud/actions/runs/23639047500/job/68854888888 ;)

MohamedAminMallek (approved):

PR #273, opened 2026-05-27
chore: upgrade min python version and packages (public-apis#1797)

Fenkiou:
### Description

We drop python 3.9 support because:
- it reaches it's EOL
- it prevents us to upgrade pulser (problematic)
- it prevents us to upgrade tenacity (minor)

As latest pulser versions requires at least python 3.10 there is no reason to keep supporting 3.9 now.

### Checklist

#### Documentation

#### Tests

- [x] Tests were run locally.

### DIFF
--- a/.github/workflows/run-tests-and-mypy-and-pre-commit-hooks.yml
+++ b/.github/workflows/run-tests-and-mypy-and-pre-commit-hooks.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.10", "3.10", "3.11", "3.12", "3.13", "3.14"]
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
 
     steps:
       - name: Checkout main code and submodules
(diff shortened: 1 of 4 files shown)


pasqal-cloud/pasqal_cloud/backends.py
```diff
@@ -26,7 +26,7 @@
 from pasqal_cloud.pasqal_cloud_connection import PasqalCloudConnection
 
 
-class RemoteEmulatorBackend(RemoteBackend, EmulatorBackend):
+class RemoteEmulatorBackend(RemoteBackend, EmulatorBackend):  # type: ignore[misc]
```
Fenkiou, line 29: mypi started to complain about config parameter not being the same type as RemoteBackend

.github/workflows/run-tests-and-mypy-and-pre-commit-hooks.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Install main dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install ./pasqal-cloud[dev] ./pulser-pasqal
+          pip install ./pasqal-cloud[dev]
```
Fenkiou, line 28: otherwise there is a dependency conflicts but pulser-pasqal shouldn't be installed with pasqal-cloud@main anyway so I removed it

MatthieuMoreau0 (approved):

PR #274, opened 2026-05-28
feat(RemoteEmulatorBackend): map job_params 'runs' to emulator 'default_num_shots' (public-apis#1861)

Fenkiou:
### Description

Previously, the 'runs' parameter on emulator backends was silently ignored with a blanket warning. This change makes 'runs' meaningful outside of an open_batch by mapping it to the EmulationConfig's 'default_num_shots' setting.

Behavior summary:
- Outside an open_batch (single job): sets default_num_shots to the provided runs value.
- Outside an open_batch (multiple jobs): sets default_num_shots to the highest runs value and warns that per-job runs is not yet supported.
- Outside an open_batch with a custom BitStrings num_shots already set: warns that runs has no effect and preserves the existing num_shots.
- Inside an open_batch: resets runs to 1 and warns, since runs has no effect on emulators in that context.

### Checklist

#### Documentation

#### Tests

- [x] Tests were run locally.

### DIFF
--- a/pasqal-cloud/pasqal_cloud/backends.py
+++ b/pasqal-cloud/pasqal_cloud/backends.py
@@ -15,6 +15,7 @@
 
 from __future__ import annotations
 
+import logging
 import warnings
 from typing import Any, ClassVar
 
@@ -24,6 +25,8 @@ from pulser.backend import BitStrings, EmulationConfig, EmulatorBackend
 from pulser.backend.remote import JobParams, RemoteBackend, RemoteResults
 
 from pasqal_cloud.pasqal_cloud_connection import PasqalCloudConnection
+
+logger = logging.getLogger(__name__)
 
 
 class RemoteEmulatorBackend(RemoteBackend, EmulatorBackend):  # type: ignore[misc]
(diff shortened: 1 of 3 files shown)


pasqal-cloud/pasqal_cloud/backends.py
```diff
@@ -82,24 +90,71 @@ def run(
...
+                        f"{self.__class__.__name__!r}'s config.",
+                        stacklevel=2,
+                    )
+
             # TODO: Stop defaulting runs=1 once it is optional on pasqal-cloud
             _job_params = [{"runs": 1, **j} for j in job_params]
+
         return super().run(job_params=_job_params, wait=wait)
 
+    def _apply_default_num_shots(self, max_runs: int) -> bool:
```
Fenkiou, line 133: For the record, this is actually something already done at a higher level here: https://github.com/pasqal-io/qoolqit/blob/main/qoolqit/execution/backends.py#L172

MohamedAminMallek:
After our discussion last week, I think the best approach is to handle this on the cluster side until we properly update the run interface and add support for emulator configuration at the job level in the cloud.

Here's what I have in mind:

* We should stop overriding the `runs` value on the SDK side.
* On the cluster side, if `runs` is set and a bitstring observable is configured with `num_shots = null`, we can automatically set `num_shots` to the value of `runs`.

@MatthieuMoreau0 @Fenkiou, what do you think?

Fenkiou:
As discussed, we don't do it this way so I'm closing it.

PR #277, opened 2026-06-04
chore: rename EmuX by RemoteXBackend

Fenkiou:
### Description

We rename all those backends to have a consistent naming with their local counterpart and prevent an issue with EmuFreeBackend renaming causing issues in https://github.com/pasqal-io/Pulser/pull/1063

### Checklist

#### Documentation

#### Tests

- [x] Tests were run locally.

### DIFF
--- a/pasqal-cloud/pasqal_cloud/backends.py
+++ b/pasqal-cloud/pasqal_cloud/backends.py
@@ -112,7 +112,7 @@ class RemoteEmulatorBackend(RemoteBackend, EmulatorBackend):  # type: ignore[mis
         )
 
 
-class EmuMPSBackend(RemoteEmulatorBackend):
+class RemoteMPSBackend(RemoteEmulatorBackend):
     """
     Backend for executing quantum programs using the EMU-MPS emulator.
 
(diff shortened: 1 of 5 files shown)


pasqal-cloud/pasqal_cloud/backends.py
```diff
@@ -161,7 +161,7 @@ class EmuSVBackend(RemoteEmulatorBackend):
     _device_type = DeviceTypeName.EMU_SV
 
 
-class EmuFreeBackend(RemoteEmulatorBackend):
+class RemoteFreeBackend(RemoteEmulatorBackend):
```
Fenkiou: Or we rename it to `RemoteQutipBackend`

MohamedAminMallek: I think `RemoteQutipBackend` is a better name, but we need to reflect the change across our other services.

We could rename it to RemoteQutipBackend and keep Free as an alias, with a warning indicating that RemoteQutipBackend should be used instead.

On the Cloud API side, once we start working on refactoring devices/backends, we can directly use RemoteQutipBackend.

MohamedAminMallek (approved):

pasqal-cloud/pasqal_cloud/backends.py
Fenkiou: I'll go with `RemoteQutipBackend` then. No need for this `RemoteFreeBackend` alias as those classes are new in `pasqal-cloud`. Let me know if I miss something.

PR #284, opened 2026-07-24
[FEAT] honor Retry-After header on HTTP 429 responses

Fenkiou:
### Description

Sleep for the server-specified duration (delay-seconds or HTTP-date) instead of the default exponential backoff when a 429 includes Retry-After.

### Checklist

#### Documentation

#### Tests

- [x] Tests were run locally.

### DIFF
--- a/pasqal-cloud/pasqal_cloud/utils/retry.py
+++ b/pasqal-cloud/pasqal_cloud/utils/retry.py
@@ -59,14 +59,11 @@ def retry_http_error(
         @functools.wraps(func)
         def wrapper(*args: Param.args, **kwargs: Param.kwargs) -> RT:
             for iteration in range(max_retries + 1):
-                print("trying")
                 # 2 seconds, 4 seconds, 8 seconds, 16 seconds, 32 seconds
                 delay = 2**iteration
                 try:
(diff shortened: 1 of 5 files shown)


pasqal-cloud/pasqal_cloud/utils/retry.py
```diff
@@ -1,14 +1,43 @@
...
 from typing_extensions import ParamSpec
 
+if TYPE_CHECKING:
+    from requests import Response
+
 Param = ParamSpec("Param")
 RT = TypeVar("RT")  # return type
 
 
+def _retry_after_seconds(response: "Response") -> int | None:
```
MatthieuMoreau0, line 18: nit: why do we typehint lazily here?

MatthieuMoreau0 (approved):
lgtm

pasqal-cloud/pasqal_cloud/utils/retry.py
Fenkiou: A bit of context here: https://peps.python.org/pep-0781/ but the rational is: "if you don't need this at runtime, only import it for type checkers to avoid circular import issues or module loading time issue".

PR #285, opened 2026-07-29
fix(http-client): extend read timeout for batch creation

Fenkiou:
### Description

Server-side processing on slow endpoints (e.g. batch creation) could
outlast the read timeout, aborting the client before the response
arrived even though the request succeeded. Callers then retried
manually, duplicating batches/jobs and burning credits.

Reverting the global timeout removal: only send_batch now gets an
extended 5min read timeout (BATCH_CREATION_TIMEOUT), other requests
keep the default 30s.

### Checklist

#### Documentation

#### Tests

- [x] Tests were run locally.

### DIFF
--- a/pasqal-cloud/pasqal_cloud/http_client.py
+++ b/pasqal-cloud/pasqal_cloud/http_client.py
@@ -262,7 +262,7 @@ class HTTPClient:
         resp = request_with_retry(
             method,
             url,
-            timeout=TIMEOUT,
+            timeout=(TIMEOUT, None),
             headers=headers,
             auth=self.authenticator,
             params=params,
--- a/pasqal-cloud/pasqal_cloud/http_client.py
+++ b/pasqal-cloud/pasqal_cloud/http_client.py
@@ -18,7 +18,7 @@ import json
 import os
 import warnings
 from getpass import getpass
-from typing import Any, Dict, List, Mapping, Optional, Sequence, Union
+from typing import Any, Dict, List, Mapping, Optional, Sequence, Tuple, Union
 from uuid import UUID
 
 import requests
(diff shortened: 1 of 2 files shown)


pasqal-cloud/pasqal_cloud/http_client.py
```diff
@@ -262,7 +262,7 @@ def _authenticated_request(
         resp = request_with_retry(
             method,
             url,
-            timeout=TIMEOUT,
+            timeout=(TIMEOUT, None),
```
MohamedAminMallek: I find this a bit risky. If the server goes down without closing the socket, we'll wait forever with no exception raised.

I'd rather set a high but finite timeout, 2 or 3 minutes.

Since the performance issue only concerns some endpoints, the ideal would be a custom timeout for those specific calls rather than for all of them. That's more work though, so in my opinion a higher timeout is a safer default for now.

What do you think?

Fenkiou: Totally agree with you, a custom timeout for batch creation would be enough for now, I'll change that

Fenkiou: done

pasqal-cloud/pasqal_cloud/http_client.py
```diff
@@ -42,6 +42,7 @@
 from pasqal_cloud.utils.retry import retry_http_error
 
 TIMEOUT = 30  # client http requests timeout after 30s
+BATCH_CREATION_TIMEOUT = 300  # batch creation can take longer, allow 5min
```
MatthieuMoreau0, line 45: Are these timings we observed in prod? 🫨

Fenkiou: It was more around a minute with a batch a 1000 jobs for a test of mine yes

MohamedAminMallek (approved):

PR #293, opened 2026-09-09
[CI] PR workflows + pin actions

sarjevane:
### Description

- Refactor workflows to run as much as possible in PRs from forks
- Update and pin all actions
- Migrate Gitleaks scan from action to binary

### Checklist

#### Documentation

#### Tests

- [ ] Tests were run locally.

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -9,22 +9,40 @@ on:
       - synchronize
       - closed
 
-concurrency: preview-${{ github.ref }}
+# `pull_request` (never `pull_request_target`) is deliberate: the build below
+# executes code from the pull request, so it must not run in a context that has
+# access to repository secrets or a writable token.
+permissions: {}
+
+concurrency:
+  group: preview-${{ github.event.pull_request.number }}
+  # Deploys push to gh-pages; cancelling one midway can leave it inconsistent.
+  cancel-in-progress: false
 
 jobs:
-  deploy-preview:
+  # UNTRUSTED. Builds the pull request's own code: `pip install .[docs]` runs
+  # its packaging code, and mkdocs.yml enables `markdown_exec`, which executes
+  # Python from the docs at build time. This job therefore holds no permissions
+  # and no credentials -- it only produces an artifact.
+  build:
+    name: Build docs
+    if: github.event.action != 'closed'
     runs-on: ubuntu-latest
+    permissions:
+      contents: read # checkout only; deliberately nothing else
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
 
-      - uses: actions/setup-python@v5
+      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
           python-version: 3.x
 
-      - run: echo "cache_id=$(date --utc '+%V')" >> $GITHUB_ENV
+      - run: echo "cache_id=$(date --utc '+%V')" >> "$GITHUB_ENV"
 
-      - uses: actions/cache@v4
+      - uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
         with:
           key: doc-publish-${{ env.cache_id }}
           path: .cache
@@ -36,8 +54,44 @@ jobs:
 
       - run: mkdocs build
 
+      - name: Upload built site
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: pr-preview-site
+          path: ./site/
+          retention-days: 1
+
+  # TRUSTED. Holds the write token but never runs code from the pull request:
+  # it checks out the base commit and only unpacks the artifact built above.
+  deploy:
+    name: Deploy preview
+    needs: [build]
+    # Forks get a read-only token no matter what `permissions` says, so a
+    # deploy from one cannot succeed. Skip it rather than fail the PR; the
+    # build job above still validates that a fork's docs compile.
+    if: |
+      always() &&
+      github.event.pull_request.head.repo.full_name == github.repository &&
+      (github.event.action == 'closed' || needs.build.result == 'success')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write       # push the preview to the gh-pages branch
+      pull-requests: write  # leave/update the sticky preview comment
+    steps:
+      - name: Checkout the base commit, never the PR head
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
+
+      - name: Download built site
+        if: github.event.action != 'closed'
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: pr-preview-site
+          path: ./site/
+
       - name: Deploy preview
-        uses: rossjrw/pr-preview-action@v1
+        uses: rossjrw/pr-preview-action@ffa7509e91a3ec8dfc2e5536c4d5c1acdf7a6de9 # v1.8.1
         with:
           source-dir: ./site/
           preview-branch: gh-pages
--- a/.github/workflows/publish-doc-to-gh-pages.yml
+++ b/.github/workflows/publish-doc-to-gh-pages.yml
@@ -9,7 +9,7 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - name: Verify tag matches VERSION.txt
         run: |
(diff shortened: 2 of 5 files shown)


.github/workflows/verify-if-secrets-were-leaked.yml
```diff
@@ -1,15 +1,79 @@
...
+  # so that a moved tag cannot silently change the code we execute.
+  GITLEAKS_IMAGE: ghcr.io/gitleaks/gitleaks@sha256:a83ee37c5eb2523689b15af8c645424c7eb1730541037e8ea805637d91372e85
 
 jobs:
   scan:
     name: gitleaks
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - name: Checkout
```
MatthieuMoreau0, line 28: Can you clarify what changed and why? Why are we no longer using the action but a docker image instead? Is it to let us scan all the commits in the history instead of just the last one?

MatthieuMoreau0 (approved):

.github/workflows/verify-if-secrets-were-leaked.yml
sarjevane: Gitleaks itself is [MIT-licensed](https://github.com/gitleaks/gitleaks?tab=MIT-1-ov-file), while the action [requires a license](https://github.com/gitleaks/gitleaks-action#environment-variables):

> `GITLEAKS_LICENSE` (required for organizations, not required for user accounts): A gitleaks-action license can be obtained at [gitleaks.io](https://gitleaks.io/). It should be added as an encrypted secret [to the repo](https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository) or [to the organization](https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-an-organization).

The issue is that we do want to run gitleak on CI for PRs from forks, but secrets (including the license) are not available to forks (they could steal/leak them otherwise), so we cannot use the license here, and we replace the implementation by manually running Gitleaks instead.

.github/workflows/preview.yml
```diff
@@ -9,22 +9,40 @@ on:
...
+    if: github.event.action != 'closed'
     runs-on: ubuntu-latest
+    permissions:
+      contents: read # checkout only; deliberately nothing else
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
```
Fenkiou, line 37: wondering why it's not the default

.github/workflows/publish-to-testpypi-and-pypi.yml
```diff
@@ -91,15 +99,17 @@ jobs:
                   }
               }
           ' CHANGELOG.md)
-          echo "RELEASE_NOTES<<EOF" >> $GITHUB_ENV
-          echo "$notes" >> $GITHUB_ENV
-          echo "EOF" >> $GITHUB_ENV
+          {
+            echo "RELEASE_NOTES<<EOF"
+            echo "$notes"
+            echo "EOF"
```
Fenkiou, line 105:
```suggestion
            echo EOF
```
?

Not sure but the doc here: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#example-of-a-multiline-string is bit a different

.github/workflows/run-tests-and-mypy-and-pre-commit-hooks.yml
```diff
@@ -1,24 +1,46 @@
...
+      - main
+      - dev
+
+# This workflow only reads the code it tests. Nothing here needs to write to
+# the repository, and no secret is referenced, so a fork pull request runs it
+# exactly as an internal branch does.
+permissions:
+  contents: read
+
+concurrency:
```
Fenkiou, line 19: Hmm, once a new commit is pushed test/lint/type check can just be cancelled as only the latest one matters no?

.github/workflows/verify-if-secrets-were-leaked.yml
```diff
@@ -1,15 +1,79 @@
...
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
-          fetch-depth: 50
-      - uses: gitleaks/gitleaks-action@v2
+          # Full history: the base commit has to be present to diff against.
+          fetch-depth: 0
+          # Never leave a usable credential in .git/config for later steps.
```
Fenkiou, line 33: Maybe comment is wrong based on https://github.com/actions/checkout#whats-new-1

> Improved credential security: persist-credentials now stores credentials in a separate file under $RUNNER_TEMP instead of directly in .git/config

Anyway, you didn't added for other actions/checkout so I'd remove it ;)

.github/workflows/publish-to-testpypi-and-pypi.yml
```diff
@@ -42,31 +48,33 @@ jobs:
...
 
   create-github-release:
     name: Create GitHub Release
     runs-on: ubuntu-latest
     needs: build-n-publish
+    permissions:
+      contents: write # required to create the GitHub release
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```
Fenkiou, line 67: forgot `persist-credentials: false` here?

Fenkiou (commented):
a bit late ahah

sarjevane:
> a bit late ahah

@Fenkiou Apologies, did not know you were reviewing! Let me address your comments, and I can open a follow-up PR with whatever we agree to fix

.github/workflows/verify-if-secrets-were-leaked.yml
sarjevane: That comment was based on the action v4 (later updated to v7), will remove to keep comments clean.

.github/workflows/run-tests-and-mypy-and-pre-commit-hooks.yml
sarjevane: As it is in this PR, in a PR I agree with you, when the commit sits in `main` or in a tag, I would leave it.

Reason behind would be if the test/lint/typecheck takes a bit longer, and we merge multiple PRs in a row, it can help pinpoint which commit broke the jobs.

We're nowhere near that kind of traffic in this repo, so I'm open to always have `cancel-in-progress: true` instead of `cancel-in-progress: PR-only`, WDYT?

.github/workflows/publish-to-testpypi-and-pypi.yml
sarjevane: `echo "EOF"` and `echo EOF` are strictly equivalent here, but I removed it in #297 with the rework of the release creation

.github/workflows/preview.yml
sarjevane: Not sure either ([source](https://github.com/actions/checkout/blob/f548e57e544e1ff5a4c46bf1e1b8685f8e4a348a/action.yml#L54)). Only guess is to be backward-compatible.

After review, there is one place where we actually do need the credentials, so I'll set that spot to `true`, all other places to `false`.

.github/workflows/publish-doc-to-gh-pages.yml
sarjevane, line 38: Note: This is actually the only place that under the hood uses git, so that's the only place where we need the `persist-credentials: true` (the default value).

.github/workflows/publish-to-testpypi-and-pypi.yml
sarjevane: Added

sarjevane (commented):
Created https://github.com/pasqal-io/pasqal-cloud/pull/297 as a follow-up.

Some questions are still open here, I can include them in #297 if you want me to:
- remove `persist-credentials` altogether
- `cancel-in-progress` for non-pr workflows

.github/workflows/run-tests-and-mypy-and-pre-commit-hooks.yml
```diff
@@ -1,24 +1,46 @@
...
-on: push
+on:
+  pull_request:
+    # `ready_for_review` matters: the job below skips drafts, and without this
+    # type a draft -> ready transition would never re-trigger the gate.
+    types: [opened, reopened, synchronize, ready_for_review]
+  push:
+    branches:
+      - main
+      - dev
```
Fenkiou, line 11: `dev` branch isn't used anymore ;)

.github/workflows/run-tests-and-mypy-and-pre-commit-hooks.yml
```diff
@@ -1,24 +1,46 @@
...
+
+# This workflow only reads the code it tests. Nothing here needs to write to
+# the repository, and no secret is referenced, so a fork pull request runs it
+# exactly as an internal branch does.
+permissions:
+  contents: read
+
+concurrency:
+  group: tests-${{ github.event.pull_request.number || github.ref }}
+  # Superseded pull request runs are pointless; runs on main/dev are not.
```
Fenkiou, line 21: `main/dev` -> `main` ;)

.github/workflows/run-tests-and-mypy-and-pre-commit-hooks.yml
Fenkiou: > can help pinpoint which commit broke the jobs

ok valid point even though I'm not sure it can happen as those jobs must pass before the PR lands in main, anyway keep it as is that's not an issue :)

PR #292, opened 2026-09-08
[FEAT] added mike and versioning to pasqal cloud

thomasbaronnet:
### Description

Versioned MkDocs documentation with mike.

On each release tag, docs are deployed as a versioned snapshot aliased as `latest`, instead of overwriting gh-pages with `mkdocs gh-deploy --force`. Documentation links now point to `/latest/`.

Preview is here https://thomasbaronnet.github.io/pasqal-cloud/latest/

### Checklist

- [x] The title of the PR follows the right format: [{Label}] {Short Message}. Label examples: IMPROVEMENT, FIX, REFACTORING... Short message is about what your PR changes.

#### Versioning (PASQAL developers only)

- [ ] Update the version of pasqal-cloud in `VERSION.txt` following the changes in your PR and by using [semantic versioning](https://semver.org/).

#### Documentation

- [ ] Update CHANGELOG.md with a description explaining briefly the changes to the users.

#### Tests

- [ ] Unit tests have been added or adjusted.
- [x] Tests were run locally.

#### Internal tests pipeline (PASQAL developers only)
- [ ] Update and run the internal tests while targeting the branch of this PR.
 If your PR hasn't changed any functionality, it still needs to be validated against internal tests.

#### After updating the version (PASQAL developers only)

- [ ] Open a PR on the internal tests that updates the version used for the pasqal-cloud backward compatibility tests.

PR #294, opened 2026-09-09
[FEAT] publishing the pasqal cloud documentation for each tag #1900

thomasbaronnet:
### Description

New PR replacing the https://github.com/pasqal-io/pasqal-cloud/pull/292

Versioned MkDocs documentation with mike.

On each release tag, docs are deployed as a versioned snapshot aliased as latest, instead of overwriting gh-pages with mkdocs gh-deploy --force. Documentation links now point to /latest/.

Preview is here https://thomasbaronnet.github.io/pasqal-cloud/latest/

### Checklist

#### Documentation

#### Tests

- [x] Tests were run locally.

### DIFF
--- a/.github/workflows/publish-doc-to-gh-pages.yml
+++ b/.github/workflows/publish-doc-to-gh-pages.yml
@@ -4,10 +4,10 @@ on:
     branches:
       - main
 concurrency:
-    group: gh-pages
-    cancel-in-progress: false
-    tags:
-      - "*"
+  group: gh-pages
+  cancel-in-progress: false
+  tags:
+    - "*"
 permissions:
   contents: write
 jobs:
@@ -15,6 +15,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-tags: true
       - name: Verify tag matches VERSION.txt
         if: startsWith(github.ref, 'refs/tags/')
         run: |
@@ -49,5 +51,10 @@ jobs:
       - name: Deploy versioned docs
         if: startsWith(github.ref, 'refs/tags/')
         run: |
-          mike deploy --push --update-aliases "${GITHUB_REF#refs/tags/}" latest
-          mike set-default --push latest
+          tag="${GITHUB_REF#refs/tags/}"
+          if [ "$tag" = "$(git tag -l '[0-9]*' --sort=-v:refname | head -n1)" ]; then
+            mike deploy --push --update-aliases "$tag" latest
+            mike set-default --push latest
+          else
+            mike deploy --push "$tag"
+          fi
--- a/.github/workflows/publish-doc-to-gh-pages.yml
+++ b/.github/workflows/publish-doc-to-gh-pages.yml
@@ -35,4 +35,8 @@ jobs:
             doc-publish-
       - run: pip install .[docs]
         working-directory: pasqal-cloud
-      - run: mkdocs gh-deploy --force
+      - name: Deploy versioned docs
+        run: |
+          git fetch origin gh-pages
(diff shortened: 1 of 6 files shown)


.github/workflows/publish-doc-to-gh-pages.yml
```diff
@@ -1,6 +1,8 @@
 name: Publish MkDocs documentation to gh-pages
 on:
   push:
+    branches:
+      - main
```
Fenkiou, line 5:
```suggestion
      - main
concurrency:
    group: gh-pages
    cancel-in-progress: false
```

otherwise merging in main and tagging right after could make jobs runs in parallel and make one fail

.github/workflows/publish-doc-to-gh-pages.yml
```diff
@@ -35,4 +38,13 @@ jobs:
...
+      - name: Fetch gh-pages
+        run: git fetch origin gh-pages --depth=1
+      - name: Deploy dev docs
+        if: github.ref == 'refs/heads/main'
+        run: mike deploy --push --update-aliases dev
+      - name: Deploy versioned docs
+        if: startsWith(github.ref, 'refs/tags/')
+        run: |
+          mike deploy --push --update-aliases "${GITHUB_REF#refs/tags/}" latest
+          mike set-default --push latest
```
Fenkiou: Not sure it will happen but if someone push a tag for an older version, let's say 0.24.1 while we're currently at 0.32.0, then 0.24.1 will become the default one and latest will point to 0.24.1

Fenkiou (changes requested):

.github/workflows/publish-doc-to-gh-pages.yml
thomasbaronnet: Yes indeed, but if we're not adding this the /latest won't work, what do you suggest instead? Is it something that might happen really?

.github/workflows/publish-doc-to-gh-pages.yml
thomasbaronnet: Ok yes I applied this

.github/workflows/publish-doc-to-gh-pages.yml
Fenkiou: isn't there a way to check if the current tag is the latest one somehow?

thomasbaronnet: we can add a `if [ "${GITHUB_REF#refs/tags/}" = "$(git tag -l '[0-9]*' --sort=-v:refname | head -n1)" ]; then` that push only if the tag is the most recent one

Fenkiou: seems like a good solution 👍 but we need to tweak a bit the `action/checkout` step in order to not fetch the whole commits history but only the highest tag (if doable or all tags)

thomasbaronnet: Ok, it's done, I added the `fetch-tags: true `on checkout so we get all tags without fetching the full commit history. Fetching only the highest tag isn't supported by checkout, so all tags it is.

.github/workflows/publish-doc-to-gh-pages.yml
```diff
@@ -1,16 +1,24 @@
...
   push:
-    tags:
-      - "*"
+    branches:
+      - main
+concurrency:
+  group: gh-pages
+  cancel-in-progress: false
+  tags:
+    - "*"
```
Fenkiou: wrong place

thomasbaronnet: Oh yeah my bad

Fenkiou (approved):

